PR #79, opened 2014-08-14
Increase control on data printed

git-harry:
maas has some particular requirements regarding the output. This
commit attempts to address the following:

- Report an error if the script attempts to return more than 30 metrics.
While the agent does log this and does stop sending metrics the status
of the check remains available. This could cause issues if the metrics
aren't used with alarms

- Uncaught exceptions generate a status message.
Exceptions don't generate a status message unless one is explicitly
set. Sometimes an exception may only occur if run by the agent. Status
messages have a limit of 256 characters so the text is taken from the
tail of the traceback.

- Log unhandled exceptions
The agent does a poor job of tracking plugin failures. This commit will
log unhandled exceptions to make debugging issues easier.

- Strip newlines from output
The agent treats each line separately. Multi-line output such as
exceptions needs to have newlines escaped so that it can be used in the
status message.

- Use the last status message
The agent uses the first status message it receives and logs any that
follow as duplicates. If an error happens after an okay status message
the message needs to be updated.

### DIFF
--- a/heat_api_local_check.py
+++ b/heat_api_local_check.py
@@ -59,9 +59,10 @@ def main(args):
 
 
 if __name__ == "__main__":
-    parser = argparse.ArgumentParser(description='Check heat API')
-    parser.add_argument('ip',
-                        type=IPv4Address,
-                        help='heat API IP address')
-    args = parser.parse_args()
-    main(args)
+    with print_output():
+        parser = argparse.ArgumentParser(description='Check heat API')
+        parser.add_argument('ip',
+                            type=IPv4Address,
+                            help='heat API IP address')
+        args = parser.parse_args()
+        main(args)
--- a/heat_api_local_check.py
+++ b/heat_api_local_check.py
@@ -18,7 +18,7 @@ import argparse
 from time import time
 from ipaddr import IPv4Address
 from maas_common import (get_auth_ref, get_heat_client, metric_bool,
-                         metric, status_ok, status_err)
+                         metric, status_ok, status_err, print_output)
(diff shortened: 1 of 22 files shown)


sigmavirus24:
Can we merge this?

odyssey4me:
LGTM, however @git-harry has indicated that this would require some small changes to the existing scripts. If those changes can be included in this PR so that it's not a breaking change then we can merge it.

git-harry:
Okay, so that should be everything to make it work. Someone else should probably run this stuff though before it's merged.

mancdaz:
@git-harry you're going to need to import the print_output function from maas_common in each of the plugins

heat_api_local_check.py
```diff
@@ -59,9 +59,10 @@ def main(args):
...
-                        help='heat API IP address')
-    args = parser.parse_args()
-    main(args)
+    with print_output():
+        parser = argparse.ArgumentParser(description='Check heat API')
+        parser.add_argument('ip',
+                            type=IPv4Address,
+                            help='heat API IP address')
+        args = parser.parse_args()
+        main(args)
```
sigmavirus24: Why not more simply do:

```py
if __name__ == "__main__":
    parser = argparse.ArgumentParser(...)
    parser.add_argument(...)
    args = parser.parse_args()
    with print_output():
        main(args)
```

The creation of the argparser and the parsing of args doesn't need to be contained in the context manager produced by print_output.

git-harry:
@mancdaz I was just testing this was reviewed :-P

PR #123, opened 2014-11-14
Cherry-pick commits from master

mattt416:
dbffa3b was never cherry-picked from master, so creating PR w/ both cherry-picked commits.  I could have created a single PR simply deleting check_maas_plugins.py, but I'd rather keep things as clean as possible.


mattt416:
@b3rnard0 We should be fine to merge this, but we now have stable/juno so these commits will need to be backported there also.

PR #144, opened 2015-01-26
Add rabbitmq channels check

mancdaz:
Check that rabbitmq connections don't have multiple channels

### DIFF
--- a/rabbitmq_status.py
+++ b/rabbitmq_status.py
@@ -23,6 +23,8 @@ from maas_common import metric, metric_bool, status_ok, status_err
 
 OVERVIEW_URL = "http://%s:%s/api/overview"
 NODES_URL = "http://%s:%s/api/nodes"
+CHANNEL_URL = "http://%s:%s/api/channels"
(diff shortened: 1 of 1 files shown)


rabbitmq_status.py
```diff
@@ -82,6 +84,20 @@ def main():
...
 
     try:
+        r = s.get(CHANNEL_URL % (options.host, options.port))
+    except requests.exceptions.ConnectionError as e:
+        status_err(str(e))
+
+    if r.ok:
+        resp_json = r.json()  # Parse the JSON once
+        if any(channel['number'] > 1 for channel in resp_json):
+            status_err('Detected RabbitMQ connections with multiple channels. Please check RabbitMQ and all Openstack consumers')
```
git-harry: This line looks like it's too long - pep8?

mancdaz:
@git-harry as discussed this is a cherry pick from master. We can fix in master and cherry pick the fix down to avoid code divergence.

Issue #142, opened 2015-01-15
CMD Monitoring


b3rnard0:
From Russel Clark: 
 

> ## Metrics without units:
> I have attached two spreadsheets.  The first (agent.plugin11_16) is an actual data pull from [see Russel]  You can see from the raw metrics that a number of the metrics don’t have a unit.  If this is something that we submit when submitting the metric, I would like to have a unit for each.  I created a pivot table so you can easily see all the metrics where unit is “unknown” and the data showing is the minimum just so you can see an example value for the metric.  I guess that technically speaking, some of these don’t have a unit, but I would like to submit a unit name that at least makes some sense.  In the second spreadsheet (Monitoring Metrics Units Questions) I have pulled all the metrics with an unknown unit and then in most cases written a proposed unit.  I have also written a question for each (many are the same)  What I would like is:

> 1. If you agree with the proposed unit, go ahead and use that or if you would recommend a different unit, please let me know.
> 2. There are a few where I don’t know what the metric is at all.  Will you give me a definition/explanation for these ones and recommend a unit?
> 3. Please will you answer the questions (Column F) – again, many are the same and I guess the answer is likely to be the same)

b3rnard0:
@andymcc and @sigmavirus24 will be helping out with this issue. 

@claco, cc'ing you so you are aware of this issue.

mancdaz:
![image](https://cloud.githubusercontent.com/assets/562262/5874533/de1b63f6-a2fa-11e4-9416-f1abded143af.png)

questions from original spreadsheet answered ^^

I will use this issue to track the addition of the agreed units to each metric.

mancdaz:
fies merged to master, stable/icehouse and stable/juno.

Corresponding fix in os-ansible-deployment pending here: 

https://bugs.launchpad.net/openstack-ansible/+bug/1415542

PR #145, opened 2015-01-26
add units to all numerical metrics

mancdaz:
Where a metric is a numerical (non boolean) value, we add a descriptive unit to be displayed with the metric

closes: #142


hughsaunders:
The units looks sensible to me.

sigmavirus24:
:+1:

PR #160, opened 2015-02-06
Use implicit truthiness to check for empty lists

git-harry:
(len(var) == 0) == not var. The second form is considered the general
pattern to use.

This commit replaces occurrences of 'len(var) == 0' with 'not var'.

### DIFF
--- a/cinder_service_check.py
+++ b/cinder_service_check.py
@@ -53,7 +53,7 @@ def check(auth_ref, args):
 
     services = r.json()['services']
 
-    if len(services) == 0:
+    if not services:
         status_err('No host(s) found in the service list')
 
     status_ok()
--- a/neutron_service_check.py
+++ b/neutron_service_check.py
@@ -36,7 +36,7 @@ def check(args):
     else:
         agents = neutron.list_agents()['agents']
 
-    if len(agents) == 0:
+    if not agents:
         status_err("No host(s) found in the agents list")
 
     # return all the things
--- a/nova_service_check.py
+++ b/nova_service_check.py
@@ -36,7 +36,7 @@ def check(args):
     else:
         services = nova.services.list()
 
-    if len(services) == 0:
+    if not services:
         status_err("No host(s) found in the service list")
 
     # return all the things


Apsu:
I disagree with this approach as it relies on implicit type coercion with a boolean operator, and muddies the waters by suggesting the integer length is a true/false value. It will work fine for this particular point in the code, but in the future if other length conditions need to be handled this particular path will have to be changed because it's the wrong type and context.

-1

git-harry:
@Apsu An empty list is an empty list and 'bool(empty_list) is False'. If the code is concerned about lists etc of particular lengths then comparisons with integers is appropriate however that is separate from it being empty. Integer length is True/False when all that matters is whether or not it is empty. This is a standard Python idiom.

Apsu:
Like I said, it will work fine in this particular point in the code, but you're coercing the type to a Bool by reducing the condition of Empty/Not-Empty to False/True, which will cease to apply if something other than Empty/Not-Empty matters to handle, which will require not just adding another condition but in fact treating the length as the integer it is, instead of the coerced boolean. Being explicit and not implicit is actually the standard Python idiom, and considered best practice.

To elaborate more fully, In Python, the following two expressions evaluate to True:
False == 0
True == 1

The following two, however, evaluate to False:
False is 0
True is 1

Because the equality operator (==) performs the implicit type coercion while the identity operator (is) does not; a Bool is not an Int. Likewise using Boolean operators (and, or, not) on Integers performs an implicit type coercion. It's not future-proofed or obvious and straightforward, particularly when there are a variety of "gotchas" around truthiness identities, conversions and coercions in Python.

For reference: http://docs.python-guide.org/en/latest/writing/style/#check-if-variable-equals-a-constant and the linked https://docs.python.org/2/library/stdtypes.html#truth-value-testing as well as basically any "best practice" document, which will suggest avoid coercions and always using explicit comparisons except when checking actual Boolean values instead of simple yes/no conditions.

dolph:
Compromise?

    not len(var)

Apsu:
@dolph You're coercing me with your coercion. I will not succumb to your managerial wiles!

git-harry:
@Apsu - as far as I can tell you dislike this patch because under different circumstances at wouldn't make sense to do it! Also, the links you provide are at odds with your position.

@dolph haha

mancdaz:
Pep8 suggests the following :

For sequences, (strings, lists, tuples), use the fact that empty sequences are false.

Yes:

if seq:
if not seq:

No:

if len(seq):
if not len(seq):

sigmavirus24:
pep8 also says "Explicit is better than implicit" i.e.,

```py
if len(seq) > 0:
   # ...

if len(seq) <= 0:
   # ...
```

over

```
if seq:
   # ...

if not seq:
   # ...
```

Apsu:
@git-harry The links certainly aren't at odds with my position. The types reference page shows the complexity of comparison operations on non-boolean types, and the style guide shows that using boolean comparisons on booleans is fine without explicit == False/True specification. The implication being that it's already explicit in the case of boolean values, and the reason they're pointing that out is because that pattern of explicit comparison is the expected best practice. Exactly like we're saying :D

git-harry:
@sigmavirus24 The Zen of Python also says "Beautiful is better than ugly" and "Readability counts" among many other things. What it isn't however is explicit in the meaning of the sentences or does it provide examples. I never see anyone writing 'return None' as an example of good explicitness. Also, I've already been throwing 'idiom' about so your welcome.

I do agree with your point about intent though, as well as doing what it's supposed to it should also be write to clearly show that.

sigmavirus24:
So "Beautiful is better than ugly" is subjective so we can ignore that entirely. And "Readability counts" backs up my point about other contributors being able to easily read the intent of the original author. Further my position is a non-position for 2 reasons:

1. I likely wrote this code
1. I am not partial to either style

git-harry:
@Apsu the style guide links the the true value testing section on the types link for a list of what is considered False which would suggest that attr just has to coerce to True or False and not be True or False.

git-harry:
@sigmavirus24 I also think the whole explicit thing is subjective or at least abused to justify needlessly verbose.

git-harry:
Who would have known this could be so contentious!

Apsu:
@git-harry @sigmavirus24 My main concern is just future-proofing (less refactoring and code-reading to understand the conditional branches) and the mental context switch of empty/non-empty -> false/true. Trust me, I've written "if not seq:" hundreds of times. Right up until someone smacked me with these two points and I realized how even with my own code I'd have to re-read it to refactor it later since I purposefully tried to be clever and type less code because I knew how the coercion would work. Thus I try to be less clever these days, and write less-capable-of-surprise code, hence my -1 :)

git-harry:
@Apsu the code is not just more concise. I would suggest that using if seq/if not seq makes for less refactoring. For example if the code was changed so that seq is given a default value of None and then the assignment of the list never happened, for whatever reason, this code would still work however using len wouldn't. Personally I also find it easier to read.

Maybe I just missing an example where this change would cause an issue. For me this code is easier to reason and more adaptable to the evolution of the code. I don't see this as trying to be clever, it's more just making use of a feature of Python.

mancdaz:
closing, because

PR #162, opened 2015-02-11
Raise an exception on invalid replication data

sigmavirus24:
Instead of returning an empty dictionary, let's raise an exception
of our own which will document (in part) why we're failing here.

### DIFF
--- a/swift-recon.py
+++ b/swift-recon.py
@@ -28,6 +28,14 @@ import re
 import subprocess
 
 
+class ParseError(maas_common.MaaSException):
+    pass
+
+
+class CommandNotRecognized(maas_common.MaaSException):
+    pass
+
+
 def recon_output(for_ring, options=None):
     """Run swift-recon and filter out extraneous printed lines.
 
@@ -79,7 +87,7 @@ def _parse_into_dict(line, parsed_by):
     if match:
         return match.groupdict()
     else:
-        return {}
+        raise ParseError("Cannot parse '{0}' for statistics.".format(line))
 
 
 def recon_stats_dicts(for_ring, options, starting_with, parsed_by):
@@ -335,10 +343,7 @@ def make_parser():
     return parser
 
 
-def main():
-    parser = make_parser()
-    args = parser.parse_args()
-
+def get_stats_from(args):
     stats = {}
     if args.recon == 'async-pendings':
         stats = swift_async()
@@ -351,9 +356,19 @@ def main():
             maas_common.status_err('no ring provided to check')
         stats = swift_replication(args.ring)
     else:
-        maas_common.status_err(
-            'unrecognized command "{0}"'.format(args.recon)
-            )
+        raise CommandNotRecognized('unrecognized command "{0}"'.format(
+            args.recon))
+    return stats
+
+
+def main():
+    parser = make_parser()
+    args = parser.parse_args()
+
+    try:
+        stats = get_stats_from(args)
+    except (ParseError, CommandNotRecognized) as e:
+        maas_common.status_error(str(e))
 
     if stats:
         maas_common.status_ok()
--- a/maas_common.py
+++ b/maas_common.py
@@ -322,6 +322,10 @@ else:
         return heat
 
 
+class MaaSException(Exception):
+    """Base MaaS plugin exception."""
+
+
 def is_token_expired(token):
     expires = datetime.datetime.strptime(token['expires'],
                                          '%Y-%m-%dT%H:%M:%SZ')


mancdaz:
@sigmavirus24 wouldn't it be better to handle this exception and pass a nice status err message, rather than letting the exception be thrown. The controlled error message can be the documentation of why we've errored out here, but it will at least be a clean error status, as opposed to the plugin failing.

So we'd do something like:

```
else:
    maas_common.status_err("Cannot parse '{0}' for statistics.".format(line))
```

Instead of

```
else:
    raise ValueError("Cannot parse '{0}' for statistics.".format(line))
````

And instead of the output being this:

```
# ./swift-recon.py --ring-type account replication
status error se_into_dict(l, parsed_by),\n  File "./swift-recon.py", line 82, in _parse_into_dict\n    raise ValueError("Cannot parse '{0}' for statistics.".format(line))\nValueError: Cannot parse '[replication_time] - No hosts returned valid data.' for statistics.\n
  File "./swift-recon.py", line 365, in <module>
    main()
  File "./swift-recon.py", line 352, in main
    stats = swift_replication(args.ring)
  File "./swift-recon.py", line 150, in swift_replication
    regexp)
  File "./swift-recon.py", line 106, in recon_stats_dicts
    recon_output(for_ring, options)))
  File "./swift-recon.py", line 104, in <lambda>
    return map(lambda l: _parse_into_dict(l, parsed_by),
  File "./swift-recon.py", line 82, in _parse_into_dict
    raise ValueError("Cannot parse '{0}' for statistics.".format(line))
ValueError: Cannot parse '[replication_time] - No hosts returned valid data.' for statistics.
```

we just see this:

```
# ./swift-recon.py --ring-type account replication
status error Cannot parse '[replication_time] - No hosts returned valid data.' for statistics.
```

**Comment by sigmavirus24** [comment]
Either one should work. I'm not opposed to either way of doing this. It seems simpler (and like a far better separation of concerns) to let the functions that exclusively control output to do just that instead of interspersing it through several functions which operate at different levels in the program. Especially if we ever start testing these scripts, having a layered approach to output seems far more logical. I'm also very much in favor of not re-raising errors caught by the `print_output` context manager and further in favor of not printing the entirety of the traceback in the `status error` line.

**Comment by mancdaz** [comment]
@sigmavirus24  Either one will of course work - the difference with forcefully throwing an exception is that no good information is passed from the agent about why it failed.  
The pattern in the rest of the plugins is that we catch any expected exceptions and pass them to the status_err function so that we handle it cleanly. The wrapper stuff catches unexpected exceptions and attempts to put them into the status_err message.  Here we are explicitly forcing the exception, and allowing it to fail the plugin itself, which is the opposite of that pattern.

I agree that we don't necessarily want to be logging an entire traceback (or at least 250 characters of it) in the message to status_err - probably a generic 'plugin has failed' with the traceback being dumped to a local log would be cleaner.

**Comment by sigmavirus24** [comment]
@mancdaz agreed. I've moved the error handling into `main()` so that `print_output` doesn't print the entire traceback and re-raise the error. That's a separate bug that should be handled in a separate PR though. =)

File: swift-recon.py [hunk 1]
```diff
@@ -351,9 +348,18 @@ def main():
...
+
+
+def main():
+    parser = make_parser()
+    args = parser.parse_args()
+
+    try:
+        stats = get_stats_from(args)
+    except (ValueError, TypeError) as e:
+        maas_common.status_error(str(e))
```
git-harry: maas_common.status_err

sigmavirus24: Why not both?

sigmavirus24: Fixed in 231c6023557f2ba4a382c2af108e13b8b0c777e4

git-harry: Both? As in status_error and status_err?

swift-recon.py
```diff
@@ -351,9 +348,18 @@ def main():
...
+    return stats
+
+
+def main():
+    parser = make_parser()
+    args = parser.parse_args()
+
+    try:
+        stats = get_stats_from(args)
+    except (ValueError, TypeError) as e:
```
git-harry: Thinking about it again, I wonder if this is a good idea. Catching built-in exceptions makes it more likely that something unintended might be caught, in the sense that str(e) might not be very specific. Maybe it would be better use status_err instead of raising exceptions.

sigmavirus24: I'll happily make all the wonky custom exceptions we could possibly want, but I'm rather convinced we shouldn't be mixing our parsing/data-fetching functions with printing data. Having all of the cases where we print data in one place makes everything a lot cleaner in the end.

git-harry: I think wonky custom exceptions are preferable to catching built-in exceptions from large parts of the code.

I think making a base exception in maas_common and using that would be sufficient. I don't think you need to create a load of specific exceptions in this file.

sigmavirus24: Fixed in  b5d7398

swift-recon.py
```diff
@@ -351,9 +356,19 @@ def main():
             maas_common.status_err('no ring provided to check')
         stats = swift_replication(args.ring)
     else:
-        maas_common.status_err(
-            'unrecognized command "{0}"'.format(args.recon)
-            )
+        raise CommandNotRecognized('unrecognized command "{0}"'.format(
```
git-harry: Just a thought but with argparse add_argument can take a choices keyword argument that will make sure args.recon is one of a sequence of values.

sigmavirus24: Yes it can, but this PR is specifically for handling bad subprocess output from swift's recon tool. We can refactor this part in a separate pull request

git-harry: You say tomato....

git-harry:
+1

mancdaz:
@sigmavirus24 could you squash these commits please?

sigmavirus24:
Oops, sorry about that.

sigmavirus24:
@mancdaz squashed.

PR #171, opened 2015-03-04
Fix swift-recon.py

### LINKED ISSUE #170: swift-recon.py plugin is broken
The swift-recon.py plugin is currently failing, due to a call to a non-existent method in maas_common:

```
root@aio1:/usr/lib/rackspace-monitoring-agent/plugins# python swift-recon.py --ring-type object replication
status error common.py", line 459, in print_output\n    yield\n  File "swift-recon.py", line 380, in <module>\n    main()\n  File "swift-recon.py", line 371, in main\n    maas_common.status_error(str(e))\nAttributeError: 'module' object has no attribute 'status_error'\n
  File "swift-recon.py", line 380, in <module>
    main()
  File "swift-recon.py", line 371, in main
    maas_common.status_error(str(e))
AttributeError: 'module' object has no attribute 'status_error'
root@aio1:/usr/lib/rackspace-monitoring-agent/plugins#
```

mattt416:
This update fixes the call to a non-existent method in maas_common
(status_error).

It does not look like this needs to be back-ported to juno

Fixes #170

### DIFF
--- a/swift-recon.py
+++ b/swift-recon.py
@@ -368,7 +368,7 @@ def main():
     try:
         stats = get_stats_from(args)
     except (ParseError, CommandNotRecognized) as e:
-        maas_common.status_error(str(e))
+        maas_common.status_err(str(e))
 
     if stats:
         maas_common.status_ok()


sigmavirus24:
:+1:

stevelle:
:+1:

PR #175, opened 2015-03-30
Galera Check not working if wsrep_local_state is != 4

BjoernT:
This pull requests adds another check if the local state is not Synced.

### DIFF
--- a/galera_check.py
+++ b/galera_check.py
@@ -106,6 +106,10 @@ def main():
             replica_status['wsrep_cluster_state_uuid']):
         status_err("the local node is out of sync")
 
+    if (int(replica_status['wsrep_local_state']) != 4 or
+            replica_status['wsrep_local_state_comment'] != "Synced"):
+        status_err("the local node is not in sync state")
+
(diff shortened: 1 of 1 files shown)


claco:
Shouldn't we really make that an else case after https://github.com/rcbops/rpc-maas/pull/175/files#diff-f5fd9e75668ee31c9857a3e2e2b7388fL111 so we catch all unknown or unplanned states?

BjoernT:
@claco  Yeah I was thinking the same. Let me lookup all status states if 4 is the only good one.

BjoernT:
According http://galeracluster.com/documentation-webpages/nodestates.html#node-state-changes, 4 is the only synced state and 5 indicated it goes into a donor state which might be still ok

mancdaz:
@BjoernT @claco since all rpc-maas code for master/kilo has been migrated to rpc-extras, could this PR please be retargeted at rpc-extras.  I'm won't migrate it myself as it appears there is still discussion around the actual implementation.

Issue #179, opened 2015-04-01
nova_api_local_check.py reports useless active/error/... metrics


BjoernT:
Currently we report metrics from the novaclient.servers.list api call but doing this for the admin tenant only. This tenant usually has no active instances or any other instances in this state so I suggest to query this information across all tenants. Pull request coming

mancdaz:
Thanks @BjoernT !!

PR #180, opened 2015-04-01
Fixing  nova_api_local_check.py reports useless active/error/... metrics...

BjoernT:
Proposed fix for https://github.com/rcbops/rpc-maas/issues/179 (nova_api_local_check.py reports useless active/error/... metrics)


mancdaz:
tested +1

Issue #163, opened 2015-02-27
Monitor connections


cemason:
Would it be possible to add a monitor to compare /proc/sys/net/ipv4/netfilter/ip_conntrack_max with /proc/sys/net/ipv4/netfilter/ip_conntrack_count and alert if the connections vs max connections is 90% or more?

mancdaz:
@cemason I've added this to our backlog, so it can be prioritised and targeted for a future sprint.

cemason:
Thanks!



—

cfarquhar:
Thanks @mancdaz and @cemason.  Please expose both metrics to MaaS so we can get them in to graphana once it's available.

cfarquhar:
Hi @mancdaz, We had another customer's environment fall over due to connections exceeding the conntrack max.  Can you please provide an update on when this will be implemented?

sigmavirus24:
@mancdaz is out of office sick, but I'll ping @b3rnard0 to get this bumped regarding priority.

b3rnard0:
@cfarquhar We are going to expedite this as part of our next sprint that starts on 4/2. We hope to have a fix some time in the early part of the sprint. I'll add an estimate closer to the 4/2 date.

@claco @mancdaz @sigmavirus24

b3rnard0:
Fixed.

PR #181, opened 2015-04-02
Add conntrack_count.py

git-harry:
This commit adds a plugin to allow the number of connections and the
maximum number allowed to be monitored.

Issue: https://github.com/rcbops/rpc-maas/issues/163


stevelle:
:+1:

claco:
-1 based on Darren's other comment here https://github.com/rcbops/rpc-extras/pull/49#issuecomment-89207780

Specifically, the original ask was for /proc/sys/net/ipv4/netfilter/ip_conntrack_count, not /proc/sys/net/netfilter/nf_conntrack_count, which seems to imply that the original request and the proposed solution are out of sync on intent somehow. Should clarify before moving forward.

cemason:
Sorry for that - the request should be for nf_conntrack_count and _max rather than ip_conntrack_count and _max.

nrb:
+1

mancdaz:
-1

I'd suggest a mention in the readme of the fact that in order to consume this plugin, one needs to ensure the nf_conntrack_ipv4 module is loaded, otherwise /proc/sys/net/netfilter/nf_conntrack_* entries will not exist. It was not loaded on the trusty image I tested this plugin on.

stevelle:
Fair point. I didn't test a stock system.

mancdaz:
@git-harry pls could you add readme note as suggested

git-harry:
@mancdaz - I've already added readme notes. What have I missed?

mancdaz:
@git-harry apologies

:+1:

mattt416:
Hi @git-harry, can you set file perms to 755?

mattt416:
@git-harry In this case it's a bit trivial, but we should probably get into the habit of feeding back the cause of the failure to maas for issues that we expect or know may happen (in this case, kernel module not being loaded and therefore the files in question not being present).  Can you update it to do that?  I'm happy otherwise and will merge once that's in.  Thanks!

git-harry:
@mattt416 - I'd considered the exception fairly self-explanatory and with the way maas_common works that automatically gets reported. However after our discussion I have added the additional error handling.

mattt416:
LGTM

PR #185, opened 2015-04-02
Update README section for service checks

### LINKED ISSUE #184: README for service checks to require hostname or ip
The README file has become out of date, since the fix for #157. Documentation still says IP is required.

stevelle:
Adds hostname as flavor of required parameter.

Fixes #184


mancdaz:
:+1:

PR #187, opened 2015-04-07
Add hp_monitoring.py script

git-harry:
The hp_monitoring.py script adds basic hardware monitoring for HP boxes
and is designed to complement the openmanage.py script for Dell
hardware.

The script makes use of the utilities hpasmcli and hpssacli. Details of
the repo from which to get them can be found at
https://downloads.linux.hp.com/SDR/project/mcp/. The packages installed
were hp-health and hpssacli.

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,3 +308,16 @@ The kernel modules nf_conntrack_ipv4 and/or nf_conntrack_ipv6 must be loaded for
     status okay
     metric nf_conntrack_max uint32 262144
     metric nf_conntrack_count uint32 354
+
+***
+#### hp_monitoring.py
+
+#### Description:
+Returns metrics indicating the status of parts of HP hardware.
+
+#### Example output:
+
+    status okay
+    metric hardware_memory_status uint32 1
+    metric hardware_processors_status uint32 1
+    metric hardware_disk_status uint32 1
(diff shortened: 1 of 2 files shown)


claco:
Should we add the installation of those packages to an hp.yml and hp_steup.yml in rpc-extras like we did for Dell?  https://github.com/rcbops/rpc-extras/tree/master/playbooks/roles/rpc_maas/tasks

git-harry:
@claco - I don't believe rpc-extras installs any apt packages.

mancdaz:
@git-harry as per the dell plugin:

https://github.com/rcbops/rpc-maas/blob/master/openmanage.py#L23

...perhaps we want a version check in here in case the output format changes between versions.  Especially given the very specific nature of some of the strings being matched - I'm looking at you 'OK)':

https://github.com/rcbops/rpc-maas/pull/187/files#diff-638695592750f3188a3ec56b99bf2cd9R38

mancdaz:
Also please add the Apache license found at the top of (most of) the other plugins (we can fix the others separately)

git-harry:
@mancdaz - I didn't add any version checking because I couldn't find a way to get a version out of hpasmcli

andymcc:
tested the hp_monitoring script, works fine - @mancdaz are we good to merge it - will require a rebase though.

mancdaz:
@andymcc yes - once rebased I'm happy for this to get merged.

git-harry:
@andymcc - done

PR #189, opened 2015-04-08
Disable warnings conditionally

sigmavirus24:
requests.packages.urllib3.disable_warnings is only available on requests
2.4.0 or later. Right now, we package have a lower limit on 2.2.0 and
all versions before 2.4.0 will cause an AttributeError when calling
this.

/cc @stevelle

### DIFF
--- a/horizon_check.py
+++ b/horizon_check.py
@@ -26,7 +26,8 @@ from lxml import html
 
 def check(args):
     # disable warning for insecure cert on horizon
-    requests.packages.urllib3.disable_warnings()
+    if requests.__build__ >= 0x020400:
+        requests.packages.urllib3.disable_warnings()
 
     splash_status_code = 0
     splash_milliseconds = 0.0


sigmavirus24:
Fun story, for some reason, I was running the horizon check against keystone which is why I saw this.

stevelle:
:+1: works in my env

mattt416:
This should have been fixed in https://review.openstack.org/#/c/166849/.

sigmavirus24:
If you build an environment with master, we get 2.2.1, which is supported in Kilo so it wont' be fixed by 166849

mattt416:
Hi @sigmavirus24 / @stevelle,

I've just deployed os-a-d from master and got:

```
root@jenk-heat-166-node1:/var/log# pip freeze | grep requests
requests==2.6.0
root@jenk-heat-166-node1:/var/log#
```

sigmavirus24:
Not sure how you got 2.6.0 from mirror.rackspace.com but I'm okay abandoning this.

mattt416:
Looks sensible to me.  

:+1:

sigmavirus24:
Counting @stevelle's +1 from earlier, I'm merging this ASAP.

Issue #190, opened 2015-04-09
Wrong permission on swift-dispersion.py / swift-recon.py


mattt416:
It looks like rpc-extras does a chmod on scripts once this repo is checked out, which means subsequent runs of setup-maas.yml will fail due to the following:

failed: [jenk-heat-167-node2] => {"failed": true}
msg: Local modifications exist in repository (force=no).

We probably need to address this on the rpc-extras side also.

PR #192, opened 2015-04-09
Ensure all scripts are executable

mattt416:
The plugins swift-dispersion.py / swift-recon.py currently have a
permission of 644, but need a permission of 755.  This commit updates
the permissions on these two files.

Closes issue #190

(cherry picked from commit ad13a6fe4c1f8a92a5e139aca7c3d271fa435249)


mattt416:
On os-a-d, it appears that we're -W juno reviews until 10.1.3 is cut.  We may want to do the same thing here.

mancdaz:
:+1:

Issue #197, opened 2015-04-16
os-a-d no longer enables nova's v3 API


mattt416:
Our nova plugins query the v3 api, however this is now disabled in os-a-d kilo and cannot be enabled.  While that functionality might be enabled, it's probably best that we use v2(.1) instead.

PR #198, opened 2015-04-16
Update nova plugins to use v2.1 API

mattt416:
Currently, these scripts use v3 API however upstream os-a-d project no
longer enables v3 API.  This change uses v2.1 API instead, which
requires including tenant_id in the URL.

Closes issue #197

### DIFF
--- a/nova_api_local_check.py
+++ b/nova_api_local_check.py
@@ -18,7 +18,7 @@ import argparse
 import collections
 from time import time
 from ipaddr import IPv4Address
-from maas_common import (get_nova_client, status_err, metric,
+from maas_common import (get_auth_ref, get_nova_client, status_err, metric,
                          status_ok, metric_bool, print_output)
 from novaclient.client import exceptions as exc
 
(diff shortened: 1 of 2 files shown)


sigmavirus24:
Perhaps a silly question, but instead of changing `check`'s signature, why not get the `auth_ref` data inside the `check` function? It results in a smaller delta too. It's not a blocker, just a question. Is there a longer term design that I haven't seen or am missing?

git-harry:
Why not allow monitoring of both? Some customer may want to continue using v3.

sigmavirus24:
I'm 98% certain that Nova's removed v3 in Kilo.

mattt416:
@sigmavirus24 no reason for doing the get_auth_ref() in main(), I've moved that out to limit the change.  Thanks!

mattt416:
@git-harry true, but we were never giving customers the option to monitor v2 or v2.1.  I don't think the plugin is there to monitor different API versions, though I do agree that could be useful for a future feature.

sigmavirus24:
:+1: from me. Looks good (although I haven't tested it against an AIO yet).

nrb:
+1.

This looks reasonable to me, but I haven't run it yet, either.

miguelgrinberg:
-1

Kinda late now that this merged, but it does not work for me, I'm getting 500s back from nova. I'll submit a patch once I figure out what's going on.

miguelgrinberg:
There are two issues that need investigation:

1. it appears that there is some ties in nova between the v3 and v2.1 APIs. If the v3 API is disabled in the configuration, the v2.1 is also skipped. Reference: https://github.com/openstack/nova/blob/97d63d8745cd9b3b391ce96b94b4da263b3a053d/nova/api/openstack/__init__.py#L291. If I enable the v3 api in the nova-api container then this works.

2. Looking at the documentation, the v2.1 API does not take tenant IDs in the URL. But in practice the requests do not work unless I include the tenant, not sure why. So I guess the documentation is wrong?

I think the safest way out of this mess is to use v2 instead of v2.1.

PR #200, opened 2015-04-17
Use nova API v2 instead of v2.1

mattt416:
In commit 347463d we changed the nova plugins to use v2.1 instead of
v3.  This was done as v3 is now deprecated and defaults to being
disabled in os-ansible-deployment.  The problem was the testing was
done on a deployment w/ v3 enabled and it appears there is some
coupling between v2.1 and v3.

This commit simply updates the nova plugins to use v2 instead of v2.1.
Thanks to @miguelgrinberg for spotting this!

Closes issue #197

### DIFF
--- a/nova_api_local_check.py
+++ b/nova_api_local_check.py
@@ -30,7 +30,7 @@ def check(args):
     auth_token = auth_ref['token']['id']
     tenant_id = auth_ref['token']['tenant']['id']
 
-    COMPUTE_ENDPOINT = 'http://{ip}:8774/v2.1/{tenant_id}' \
+    COMPUTE_ENDPOINT = 'http://{ip}:8774/v2/{tenant_id}' \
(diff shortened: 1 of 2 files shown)


mattt416:
@miguelgrinberg left some good feedback on this issue in https://github.com/rcbops/rpc-maas/pull/198#issuecomment-93867002.

odyssey4me:
While I haven't tested this, I know that @mattt416 and @miguelgrinberg have put a considerable amount of time into it. While the more desirable option was to go with the v2.1 API, this has turned out to be more complicated than originally thought as it seems that the upstream work on it may be incomplete or badly documented. I agree with the approach of falling back to the v2 API in the spirit of keeping things simple and resolving this in time to meet our other goals.

miguelgrinberg:
:+1:

I think this is the right choice for maas.

nrb:
Was going to leave a +1 but Miguel merged already :)

Issue #188, opened 2015-04-08
rabbitmq_status.py issues


mattt416:
I've started to see this type of failure on newly built clusters:

```
root@jenk-heat-158-node1:/usr/lib/rackspace-monitoring-agent/plugins# ./rabbitmq_status.py -H 172.29.236.126 -n jenk-heat-158-node1_rabbit_mq_container-fd99ca31
status error Detected RabbitMQ connections with multiple channels. Please check RabbitMQ and all Openstack consumers
root@jenk-heat-158-node1:/usr/lib/rackspace-monitoring-agent/plugins#
```

It looks like https://github.com/rcbops/rpc-maas/pull/139 introduced this change -- however the issue seems to involve us alarming when `channel['number'] > 1`, but `number` here is actually the channel number which doesn't make sense to alarm on (assuming I'm reading this correctly).

I believe what @cfarquhar requested was not implemented -- we should be checking /api/connections and not /api/channels, and perhaps we can alarm if channels > channel_max?  Some research will need to be done here as I'm not overly clear on the specifics of rabbitmq channels.

Mentioning @BjoernT for input.

--Matt

mattt416:
Additionally, it would probably make more sense to write a metric for whatever we are testing here and then alarm against that, rather than just bail the plugin if some value is too high.  The problem w/ doing the latter is that we end up getting alerts for this that say rabbitmq_disk_free_alarm_status which is a bit misleading if you don't dig into the details of the alert.

BjoernT:
After all the rule should be still, one connection has one channel. There is a blueprint for openstack to reuse connection while opening multiple channels. I do not know to which  RPC/openstack version this issue is related to, which might explain the change in behavior

b3rnard0:
@mattt416 Please look into this though it's still a low priority issue.

mattt416:
This is still not solved -- it doesn't fail constantly now but I do still get periodic errors on a relatively unused cluster.

nrb:
Can this be closed? It appears to be merged in kilo and juno.

nrb:
Actually I'm just going to close it, and if it shouldn't be, please re-open.

PR #203, opened 2015-05-06
Update rabbitmq_status.py

mattt416:
This commit updates rabbitmq_status.py to check /api/connections
instead of /api/channels.  According to rabbitmqctl, 'numbers'
on /api/channels is:

The number of the channel, which uniquely identifies it within a
connection.

Alarming if number > 1 is ineffective, and on OpenStack Kilo this
actually poses a problem as doing a 'rabbitmqctl list_channels
numbers' actually returns channels w/ name > 1.

Further, in https://bugs.launchpad.net/oslo.messaging/+bug/1406629 we
see the following command being used to determine problem state:

rabbitmqctl list_connections name channels | awk '{ if ($4 >1) print }'

This confirms that we should be checking /api/connections and not
/api/channels.

Closes issues #188

### DIFF
--- a/rabbitmq_status.py
+++ b/rabbitmq_status.py
@@ -24,7 +24,7 @@ from maas_common import (metric, metric_bool, status_ok, status_err,
 
 OVERVIEW_URL = "http://%s:%s/api/overview"
 NODES_URL = "http://%s:%s/api/nodes"
-CHANNEL_URL = "http://%s:%s/api/channels"
+CONNECTIONS_URL = "http://%s:%s/api/connections"
 
 CLUSTERED = True
 CLUSTER_SIZE = 3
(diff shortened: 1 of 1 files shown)


mattt416:
Hi @BjoernT,

Can you please review this since you originally implemented this functionality?

Thanks!

--Matt

git-harry:
+1

BjoernT:
Yes it should be alright, especially for Kilo.

sigmavirus24:
LGTM :+1:

BjoernT:
@sigmavirus24 Can we get this merged into the kilo branch too ?

sigmavirus24:
@BjoernT see #207

BjoernT:
:+1:

PR #206, opened 2015-05-19
Create rabbitmq_max_channels_per_conn metric

mattt416:
Currently, we alarm if connection['channels'] > 1.  This change creates
a new metric called rabbitmq_max_channels_per_conn which we can alarm
against using a MaaS alarm.  Additionally, this change updates
CONNECTIONS_URL so we only return the data from /api/connections that
we need (channels).

TODO: We will need to create an alarm on rpc-extras which ideally
      alarms if rabbitmq_max_channels_per_conn > 10.  Running a
      dev cluster periodically sees rabbitmq_max_channels_per_conn > 1
      and 10 is an 'arbitrary but likely reasonable limit' which we can
      use to prevent channel ballooning.

Closes issue #188

### DIFF
--- a/rabbitmq_status.py
+++ b/rabbitmq_status.py
@@ -24,7 +24,7 @@ from maas_common import (metric, metric_bool, status_ok, status_err,
 
 OVERVIEW_URL = "http://%s:%s/api/overview"
 NODES_URL = "http://%s:%s/api/nodes"
-CONNECTIONS_URL = "http://%s:%s/api/connections"
+CONNECTIONS_URL = "http://%s:%s/api/connections?columns=channels"
 
 CLUSTERED = True
 CLUSTER_SIZE = 3
@@ -50,6 +50,8 @@ NODES_METRICS = {"proc_used": "processes",
                  "mem_alarm": "status",
                  "disk_free_alarm": "status",
                  "uptime": "ms"}
+
+CONNECTIONS_METRICS = {"max_channels_per_conn": "channels"}
 
 
 def hostname():
@@ -93,12 +95,12 @@ def main():
 
     if r.ok:
         resp_json = r.json()  # Parse the JSON once
-        if any('channels' in connection and connection['channels'] > 1 for connection in resp_json):
-            status_err('Detected RabbitMQ connections with multiple channels. Please check RabbitMQ and all Openstack consumers')
+        max_chans = max(connection['channels'] for connection in resp_json if 'channels' in connection)
+        for k in CONNECTIONS_METRICS:
+            metrics[k] = {'value': max_chans, 'unit': CONNECTIONS_METRICS[k]}
     else:
         status_err('Received status {0} from RabbitMQ API'.format(
             r.status_code))
-
 
     try:
         r = s.get(OVERVIEW_URL % (options.host, options.port))
@@ -149,8 +151,6 @@ def main():
             status_err('cluster too small')
         if not is_cluster_member:
             status_err('{0} not a member of the cluster'.format(name))
-
-
 
     status_ok()
 


rabbitmq_status.py
```diff
@@ -93,8 +93,15 @@ def main():
 
     if r.ok:
         resp_json = r.json()  # Parse the JSON once
-        if any('channels' in connection and connection['channels'] > 1 for connection in resp_json):
-            status_err('Detected RabbitMQ connections with multiple channels. Please check RabbitMQ and all Openstack consumers')
+        # We routinely see connection['channels'] == 2, so we cannot alarm if
+        # connection['channels'] > 1.  Note that connection['channels'] does
+        # not appear to be the total number of channels using the connection,
+        # but rather the channel number which identifies it within a
+        # connection.  We'll assume if we have a channel w/ number 11 or
```
git-harry: I'm a bit confused here, https://github.com/rcbops/rpc-maas/commit/fe61c70666693f14eb9fcc92431c49f278b60068, was made because the check was being done against the channel number. It now sounds like connection['channels'] does the same thing as channel['number'] in the previous commit which seems strange.

Assuming that connection['channels'] is the channel number what is the significance of the number being > 10?

Apsu: A connection can have multiple channels, and each channel has a "number", which is its index in the channel list for that connection, starting at 1. When querying the management API from the channel perspective (channel["number"]), we see the index of the channel being queried; from the connection perspective (connection["channels"]), we see the count of channels for that connection.

As @mattt416 and I discovered today, it's fairly common for there to be a couple of channels (at least briefly) for openstack component connections. The original concern Support had was with broken rabbit/clients causing ever-increasing channel creation, and the monitor was intending to catch that by making sure there was never a channel with an index ("number") > 1. So, instead, this change is attempting to put an arbitrary but likely reasonable limit on channel ballooning, without assuming that component connections will only ever use a single channel (which they demonstrably do not).

git-harry: @Apsu - thanks for the further clarification. Unless I'm misunderstanding you it sounds like you are contradicting the comment in this patch, specifically:

"Note that connection['channels'] does
+        # not appear to be the total number of channels using the connection,
+        # but rather the channel number which identifies it within a
+        # connection."

@mattt416 - had you considered making that value a metric and creating a maas alarm for it if the number is > 10?

I think for future reference it would be worth including a bit more detail in a comment/commit message for this error so that it is clear what 10 represents, why it was chosen.

mattt416:
Hi @git-harry, the reason why I stated that I believe connection['channels'] does not appear to be the total # of channels using the connection is because I was seeing different results when querying /api/connections?columns=channels versus doing a `rabbitmqctl list_connections channels` (the former would periodically return a connection['channels'] == 2 but I was never able to observe this via rabbitmqctl.  This observation may seem to invalidate the need for the commit in https://github.com/rcbops/rpc-maas/commit/fe61c70666693f14eb9fcc92431c49f278b60068 (which was implemented to mimic `rabbitmqctl list_connections channels` ("Number of channels using the connection." from rabbitmqctl's man page)), however keep in mind that the original channel which queried /api/channels was throwing an error if a channel w/ number > 1 existed, and not if a connection was using actually using that channel.  I am open to any suggestions as to how we should approach this here as there is no conclusive documentation on /api that I have been able to find.

Regarding adding a metric -- yes, I did consider doing that but my primary concern was to fix the issue I found which was triggering a failure on _every_ run on the plugin.

git-harry:
@mattt416 - is there a mailing list or IRC channel where we can get the meaning of these numbers confirmed? My concern is that we aren't sure if these numbers actually tell us what we want to know.

Moving the condition to the alarm means it's not hardcoded into the script. It also allows us to gather data to track these numbers and see if they tell us anything useful.

mattt416:
@git-harry Verified with the following code that connection['channels'] is in fact the number of channels using a connection:

```
#!/usr/bin/env python
import pika

connection = pika.BlockingConnection(pika.ConnectionParameters(host='localhost'))
channel = connection.channel()
channel2 = connection.channel()
channel3 = connection.channel(channel_number=10)

channel.queue_declare(queue='hello')

print ' [*] Waiting for messages. To exit press CTRL+C'

def callback(ch, method, properties, body):
    print " [x] Received %r" % (body,)

channel.basic_consume(callback,
                      queue='hello',
                      no_ack=True)

channel.start_consuming()
```

I'm still unsure why I'm not seeing the same results via rabbitmqctl and /api/connections however I am confident that what we are querying here is what we want.

Apsu:
@mattt416 LGTM

claco:
@mattt416 Needs backport to Juno?

mattt416:
@claco This probably isn't necessary for juno, but there are a series of commits that @BjoernT would probably like to go into kilo.  I can get those queued up if necessary.

mattt416:
This needs backporting to kilo, not juno.  The commits which this commit builds upon have not been backported to juno.

claco:
@mattt416 Yes. That was my next question. Shouldn't they have been, since these are monitoring asks from installations, that will have Juno for another 6-7 months?

d34dh0r53:
This looks good to me

cloudnull:
:+1: but someone else will have to merge.

stevelle:
this should happen in rcbops/rpc-extras repo now

mattt416:
Moved to https://github.com/rcbops/rpc-extras/pull/134.

PR #207, opened 2015-05-26
Update rabbitmq_status.py

sigmavirus24:
This commit updates rabbitmq_status.py to check /api/connections
instead of /api/channels.  According to rabbitmqctl, 'numbers'
on /api/channels is:

The number of the channel, which uniquely identifies it within a
connection.

Alarming if number > 1 is ineffective, and on OpenStack Kilo this
actually poses a problem as doing a 'rabbitmqctl list_channels
numbers' actually returns channels w/ name > 1.

Further, in https://bugs.launchpad.net/oslo.messaging/+bug/1406629 we
see the following command being used to determine problem state:

rabbitmqctl list_connections name channels | awk '{ if ($4 >1) print }'

This confirms that we should be checking /api/connections and not
/api/channels.

Closes issues #188

(cherry picked from commit de6756dbf5ea61b8c0478aa67736df81c7e89212)

/cc @BjoernT @mattt416

### DIFF
--- a/rabbitmq_status.py
+++ b/rabbitmq_status.py
@@ -24,7 +24,7 @@ from maas_common import (metric, metric_bool, status_ok, status_err,
 
 OVERVIEW_URL = "http://%s:%s/api/overview"
 NODES_URL = "http://%s:%s/api/nodes"
-CHANNEL_URL = "http://%s:%s/api/channels"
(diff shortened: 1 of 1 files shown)


mattt416:
:+1:

mattt416:
Actually, closing this in favour of https://github.com/rcbops/rpc-maas/pull/208.

PR #217, opened 2015-08-17
cinder_service_check.py not using --host option

claco:
The script cinder_service_check.py defines an optional host parameter
that is designed to filter the results, returned by the cinder API, so
that only services on the specified host are returned. The MaaS setup
role, that configures the checks that use this script, uses '--host'.
The script does not use the host argument if supplied.

This patch fixes the script so that the results returned are filtered
by host, where host includes any backends/pools on that host.

Closes-bug: #216

### DIFF
--- a/cinder_service_check.py
+++ b/cinder_service_check.py
@@ -53,6 +53,12 @@ def check(auth_ref, args):
 
     services = r.json()['services']
 
+    # We need to match against a host of X and X@lvm (or whatever backend)
+    if args.host:
+        backend = ''.join((args.host, '@'))
+        services = [s for s in services if s['host'].startswith(backend) or
+                                           s['host'] == args.host]
+
     if len(services) == 0:
         status_err('No host(s) found in the service list')
 
@@ -62,8 +68,7 @@ def check(auth_ref, args):
         if service['status'] == 'enabled' and service['state'] != 'up':
             service_is_up = False
 
-        # We need to match against a host of X and X@lvm (or whatever backend)
-        if args.host and args.host in service['host']:
+        if args.host:
             name = '%s_status' % service['binary']
         else:
             name = '%s_on_host_%s' % (service['binary'], service['host'])


cinder_service_check.py
```diff
@@ -53,6 +53,12 @@ def check(auth_ref, args):
 
     services = r.json()['services']
 
+    # We need to match against a host of X and X@lvm (or whatever backend)
+    if args.host:
+        backend = ''.join((args.host, '@'))
```
sigmavirus24: I know this is a backport but who let this through?

sigmavirus24:
:cry: :+1:

cinder_service_check.py
claco: The answer is you. :-P

sigmavirus24: Actually I didn't review this otherwise I would have brought it up in the original review.

git-harry: @sigmavirus24 I don't understand your comment, what is it you dislike about the patch?

sigmavirus24: Using `str.join` for two items. It's vastly more performent for an indeterminate number of strings, but not when you're using just 2. Doing `args.host + '@'` would be much better in this case.

git-harry: I don't think it's that clear cut. I doubt it makes any real difference performance wise which makes it a discussion on style and I like the consistency of always doing it that way.

sigmavirus24:
```
~ python -m timeit -n 1000000 '"".join(("foo", "bar"))'
1000000 loops, best of 3: 0.104 usec per loop
~ python -m timeit -n 1000000 '"foo" + "bar"'
1000000 loops, best of 3: 0.0172 usec per loop
```

There is a difference of 10x performance wise. It isn't significant here, but consistency is the hobgoblin of foolish minds and we should choose the best solution, not the most consistent.

git-harry: Best doesn't necessarily mean fastest. If Python suddenly cares about those sorts of numbers everything goes out the window.

prometheanfire:
:+1:

PR #220, opened 2015-10-02
Add maas external check for neutron metadata proxy

prometheanfire:
The neutron metadata agent proxy provides access from within neutron
network namespaces to the neutron metadata agent and thus to the
various sources of metadata provided to instances on 169.254.169.254
at port 80 when cloud-init is used.

The metadata agent has monitoring in RPC but has been reported to be
running but non-responsive on some occasions (see issue #342). The
new check tests the agent's proxy from within the namespace to
monitor responsiveness. Because the agent will use the client host
IP to look for metadata but the request is not originating from an
instance a 404 Not Found response is expected from the agent.

Because local checks are expected to operate from the host rather
than inside a container, and the host does not have ready access to
neutron network namespaces, the monitor must probe for an appropriate
container and resorts to executing a curl request through the network
namespace within the selected container for each network that is
expected to be proxied (only those networks which have ports). This is
deemed easier to verify and understand through command shell
operations than alternative solutions.

This solution seems to work regardless of the value of OSA's
nova_force_config_drive setting or the use of the config-drive feature
in Nova.

### DIFF
--- a/neutron_metadata_local_check.py
+++ b/neutron_metadata_local_check.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python
+
+# Copyright 2014, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import shlex
+import subprocess
+
+from maas_common import get_neutron_client
+from maas_common import metric_bool
+from maas_common import print_output
+from maas_common import status_err
+from maas_common import status_ok
+
+# identify the first active neutron agents container on this host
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Is this a cherry-pick or is this new development work?

prometheanfire:
it's a cherry pick (as much as you can for maas checks to juno

sigmavirus24:
Please update the commit message appropriately to include the (git generated) line to include that as we've asked you to do several times before.

:-1:

prometheanfire:
this is not a backport within this repo, it's sourced from another one

prometheanfire:
@sigmavirus24 how should I get this from https://github.com/rcbops/rpc-openstack/commit/3c598136208610e99e65d2ef495925c4ade209de because I can only do part of it, in a totally different location.  This is not a true cherry-pick because it can't be.

prometheanfire:
anyway, I'm blocked on this https://bugs.launchpad.net/openstack-ansible/+bug/1502240 as I will need to update the sha pointer in the juno branch of osad after this merges

stevelle:
mentioning it is a backport, and identifying the source repo and sha in the commit msg would go a long way toward helping make the connection clearer. It won't be a cherry-pick, because nothing maas related can be.

prometheanfire:
done

stevelle:
thanks for backporting, btw

prometheanfire:
np, it's the best way to spend my friday :P

d34dh0r53:
:+1:

sigmavirus24:
:+1:

PR #223, opened 2015-11-06
synchronize kilo branch with all plugins in rpc-openstack

mancdaz:
We have a requirement to move the plugins back out to rpc-maas, in
order that other groups can consume (and contribute to) the plugins
more easily.

This commit synchronises the plugins here with those that exist inside
rpc-openstack at this time.

Note - there is a funky way of preserving all history of commits to
these files when they were in rpc-openstack (using git filter-branch).
Sadly, there is very little history to preserve since we lost it in
the commit which moved all plugins to the plugins directory (7a3cd17)
because 'git mv' was not used from what I can tell. This means that I
can't cleanly apply that history here, and so a big commit including
all the changes is the only way to go. Sorry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,9 @@ The contents of this file would look something like this:
 
 ##### Description:
 Polls a nova API living on the specified IP. Checks to see if the API is up, and then also gathers a list of metrics to return
-##### mandatory args:
+##### Mandatory Arguments:
 IP address of service to test
-##### example output:
+##### Example Output:
 
     status okay
     metric nova_api_local_status uint32 1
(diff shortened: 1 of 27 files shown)


d34dh0r53:
:+1: this looks good to me, and I can see why we need to do this.  It's going to be a bit of a bummer to carry another submodule but there isn't really a way around it.

prometheanfire:
It looks ok but I just want to make sure you pulled from kilo, also, what are we going to do about the master stuff.  IIRC the recent work of the mass stuff was in master only.

d34dh0r53:
Good point @prometheanfire I just assumed that this incorporated the refactor.

mancdaz:
What work?  This is just the plugins - the refactor doesn't affect them - it's simply a different way to configure the checks/alarms.

Yes, this was all synchronised from the kilo branch of rpc-openstack, but I don't believe there's any difference in the plugins between master and kilo.

hughsaunders:
LGTM, checked out and all the files match except galera_check because there are PRs open against rpc-openstack that haven't yet merged. https://github.com/rcbops/rpc-openstack/pull/551 https://github.com/rcbops/rpc-openstack/pull/552

PR #224, opened 2015-11-09
elasticsearch query monitor

miguelgrinberg:
This is a simple monitor script that runs the provided elasticsearch query and reports how many hits were found.

### DIFF
--- a/elasticsearch_query.py
+++ b/elasticsearch_query.py
@@ -14,11 +14,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import datetime
 import json
 import optparse
 import os
 import re
 import requests
+import sys
 
 from maas_common import metric, status_ok, status_err, print_output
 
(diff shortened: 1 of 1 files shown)


elasticsearch_query.py
```diff
@@ -0,0 +1,166 @@
...
+        return ELASTICSEARCH + '/' + index + '/_count'
+    return ELASTICSEARCH + '/_count'
+
+
+def get_json(url, data):
+    """Wrap calls to requests to handle exceptions."""
+    exceptions = (requests.exceptions.HTTPError,
+                  requests.exceptions.ConnectionError)
+    try:
+        r = requests.get(url, data=data)
```
sigmavirus24: Does Elasticsearch return JSON even during a 404 or 411 or something else? If not you might want to add (to this try block)

```py
r.raise_for_status()
```

Which will raise an HTTPError.

sigmavirus24:
One note of interest, otherwise :+1:

elasticsearch_query.py
```diff
@@ -0,0 +1,166 @@
...
+    """Retrieve the number of hits for a query."""
+    url = count_url_for(index)
+    data = json_querystring({'query': query})
+    json = get_json(url, data)
+    return json['count']
+
+
+def parse_args():
+    parser = optparse.OptionParser(usage=('%prog [-h] [-H host] [-P port] '
+                                          '[-q query] [-i interval] '
```
mancdaz: I'm not sure if this is holy war territory, but why optparse rather than argparse? We seem to have a mix amongst our plugins and I wondered if it might be an idea to standardise on one or the other. Or doesn't anyone really care :) ?

miguelgrinberg: For the record, I'm argparse all the way. But I took the elasticsearch.py script from this repo as a base for this, so I kept optparse to be consistent.

elasticsearch_query.py
miguelgrinberg: Thanks. I think it's better to be safe and raise if there is an error, I'll make the change.

elasticsearch_query.py
stevelle: I'd file a bug as debt to make these all argparse for consistency, except I can't decide which repo to create it in.

sigmavirus24:
I'm :+1: on this regardless of `optparse` vs `argparse` because it's functional and will provide value either way.

We can refactor everything to use `argparse` down the line, but this should go in as is.

@stevelle thoughts?

stevelle:
@sigmavirus24 agreed on this, but now we will need to port this to rpc-openstack

PR #226, opened 2015-12-01
Add some connection related metrics to galera_check plugin

mancdaz:
_NOTE: this is a partial cherry-pick (essentially without the template bits, and only the plugin bits) of the following (already merged) commit from master:_

_https://github.com/rcbops/rpc-openstack/commit/5b24196daa6112f711912bea928542f9ae0a23c9_

---
add some connection related metrics to galera_check plugin

Up to now, we've been grabbing specific matching data from the output
of 'SHOW STATUS'. This commit adds the output from 'SHOW VARIABLES', so
that we can grab the configuration vars too. Additionally, we no longer
limit the captured data from 'SHOW STATUS'.

Having all of the output from 'SHOW VARIABLES' and 'SHOW STATUS' will
make it easy to add any further metrics/alarms to this plugin in the
future.

Additionally, 'SHOW STATUS' was changed to 'SHOW GLOBAL STATUS' to
ensure we are gathering global stats as opposed to 'this session' stats.

Some new metrics are also added:

current_connections
max_configured_connections
max_seen_connections

Partially addresses: https://github.com/rcbops/rpc-openstack/issues/550

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,9 @@ connects to an individual member of a galera cluster and checks various statuses
     metric wsrep_cluster_status string primary
     metric wsrep_local_state_uuid string 67e41d08-165d-11e4-9d87-7e94ef43b302
     metric wsrep_local_state_comment string synced
+    metric mysql_max_configured_connections int64 800 connections
+    metric mysql_current_connections int64 1 connections
+    metric mysql_max_seen_connections int64 2 connections
 
 ***
 #### hp_monitoring.py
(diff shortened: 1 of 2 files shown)


stevelle:
The backport is clean and there shouldn't be major issues coming from a Kilo-centric view. I haven't put together a juno env to run it yet but soft +1

sigmavirus24:
Soft +1 from me as well.

andymcc:
I'm HARD yes'ing this +2! BOOM! MERGE IT.

PR #232, opened 2017-04-26
Set the vars to proper defaults

evrardjp:
RPC-Openstack sets ssl_check to false in user variables and this
was never tested in gating with a different variable.

Let's adapt the default, and if needed later, increase test
coverage by adding a scenario to test this.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -88,7 +88,7 @@ maas_target_alias: public0_v4
 #
 # ssl_check: This flag indicates if the ssl-related checks and alarms should be deployed.
 #
-ssl_check: true
+ssl_check: false
 
 #
 # host_check: This flag indicates if the host-related hardware monitoring checks and alarms should


evrardjp:
Because that's the defaults in user requirements for the sha.

evrardjp:
On top of that, changes won't be taken into consideration until sha bumped into RPC-O, which should give enough time to understand.

alextricity25:
> On top of that, changes won't be taken into consideration until sha bumped into RPC-O, which should give enough time to understand.

Ah, right! Thanks!

I still think we should figure out how to tests these changes before we merge *anything*. If we can't do it by the time we get more direction regarding releases, then maybe we should revert the changes to remove the maas role from rpc-openstack.

evrardjp:
The only problem we could technically have, is that if there is another consumer than rpc-openstack that doesn't pin this repo. But it's not a problem OUR code can solve.

evrardjp:
@alextricity25 I see other similar changes, I will put them in different PRs. When we'll have a decision and tests, we can progressively merge this.

cloudnull:
recheck_cit_all

PR #233, opened 2017-04-26
Allow the use of FQDNs in neutron_service_check.py

cjloader:
This change modifies the neutron_service_check.py plugin to
utilize the FQDN as an argument for checking neutron services.
It was previously only using the hostname.

Connects rcbops/rpc-openstack#1603


cloudnull:
restarting the gate jobs

cloudnull:
@cjloader Initial look seems fine. Please include a release note regarding the change.

cjloader:
@cloudnull releasenote added

PR #239, opened 2017-05-03
Restructured the RPC-MAAS role

cloudnull:
This change restructures the rpc-maas tasks so that there's reduced
duplication and a lot more clarity regarding the tasks being executed.
The task files are all now prefixed with "maas_" and grouped based on
their various concerns.

This also brings with it the first set of dynamically generated checks
which is starting with the basic CDM checks. These checks have been
moved into the vars/main.yml file. This removes the template sprawl we
have and allows checks to be generated with a common core.


mkam:
recheck_cit_all

cloudnull:
recheck_cit_all

cloudnull:
@prometheanfire not likely to be able to split it into multiple reviews, the nature of this change is to pull the role forward using ansible best practices and logical task names. Shimming this change into the old structure for more incremental change won't be worth the effort.

prometheanfire:
sure, but I don't have to be happy about it.

cloudnull:
This is true! I appreciate your unhappy review.

prometheanfire:
best type of review :D

d34dh0r53:
Looks good to me!

PR #240, opened 2017-05-04
Add additional functional testing with haproxy

cloudnull:
This change allows the testing infra to begin verifying the checks deployed. To do this we need to have haproxy running because maas has a hard dependency on load balancer based monitoring at the moment.

### DIFF
--- a/tests/ansible-role-requirements.yml
+++ b/tests/ansible-role-requirements.yml
@@ -79,7 +79,6 @@
   src: https://git.openstack.org/openstack/openstack-ansible-plugins
   scm: git
   version: stable/newton
-  version: stable/newton
 - name: haproxy_server
   src: https://git.openstack.org/openstack/openstack-ansible-haproxy_server
   scm: git
--- a/tests/ansible-role-requirements.yml
+++ b/tests/ansible-role-requirements.yml
@@ -79,3 +79,8 @@
   src: https://git.openstack.org/openstack/openstack-ansible-plugins
   scm: git
   version: stable/newton
+  version: stable/newton
+- name: haproxy_server
+  src: https://git.openstack.org/openstack/openstack-ansible-haproxy_server
+  scm: git
+  version: stable/newton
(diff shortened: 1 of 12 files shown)


tests/ansible-role-requirements.yml
```diff
@@ -79,3 +79,8 @@
   src: https://git.openstack.org/openstack/openstack-ansible-plugins
   scm: git
   version: stable/newton
+  version: stable/newton
```
prometheanfire: double version?

tests/rpc-maas-overrides.yml
```diff
@@ -31,10 +31,12 @@ bridges:
 
 
 # Set the vip address
-external_lb_vip_address: 127.0.0.1
-internal_lb_vip_address: 127.0.0.1
+external_lb_vip_address: 10.1.1.51
+internal_lb_vip_address: 10.2.1.51
 internal_vip_address: "{{ internal_lb_vip_address }}"
 
+# Define the galera monitoring user
```
prometheanfire: this isn't haproxy?

cloudnull: no. this is the user defined in the galera_server role for use with monitoring.

tests/ansible-role-requirements.yml
cloudnull: done.

tests/rpc-maas-overrides.yml
prometheanfire: just another fix that snuck in?

cloudnull: No, the "galera_server" role creates the user and in order for our "haproxy_server" role to allow traffic to the galera endpoint we have to have the monitoring user defined and provided into the haproxy config. I could just hard code it, but I put it here for greater transparency. Maybe I should?

prometheanfire: no, that's good

PR #241, opened 2017-05-04
Adds consumers metrics to maas plugin

dankolbrs:
Adds metrics to rpc maas plugin for consumers. Provides total
consumer count, as well as consumer count for each vhost.

connects rcbops/u-suk-dev#855

### DIFF
--- a/files/plugins/rabbitmq_status.py
+++ b/files/plugins/rabbitmq_status.py
@@ -29,6 +29,8 @@ OVERVIEW_URL = "http://%s:%s/api/overview"
 NODES_URL = "http://%s:%s/api/nodes"
 CONNECTIONS_URL = "http://%s:%s/api/connections?columns=channels"
 QUEUES_URL = "http://%s:%s/api/queues"
+CONSUMERS_QUEUES_URL = "http://%s:%s/api/queues/%s"
+VHOSTS_URL = "http://%s:%s/api/vhosts"
 
 CLUSTERED = True
 CLUSTER_SIZE = 3
@@ -172,6 +174,29 @@ def _get_queue_metrics(session, metrics, host, port):
     }
 
 
+def _get_consumer_metrics(session, metrics, host, port):
+    VHOSTS = []
+    vhosts_response = _get_rabbit_json(session, VHOSTS_URL % (host, port))
+    # get the vhosts
+    for vhost in vhosts_response:
+        VHOSTS.append(vhost['name'])
+
+    queues_without_consumers = 0
+    for vhost in VHOSTS:
+        # uri encode /
+        URI_ENCODED = vhost.replace('/', '%2F')
+        queue_response = _get_rabbit_json(session, CONSUMERS_QUEUES_URL %
+                                          (host, port, URI_ENCODED))
+        for queue in queue_response:
+            if queue['consumers'] == 0:
+                queues_without_consumers += 1
+
+    metrics['queues_without_consumers'] = {
+        'value': queues_without_consumers,
+        'unit': 'queues'
+    }
+
+
 def main():
     (options, _) = parse_args()
     metrics = {}
@@ -183,6 +208,7 @@ def main():
     _get_node_metrics(session, metrics, options.host, options.port,
                       options.name)
     _get_queue_metrics(session, metrics, options.host, options.port)
+    _get_consumer_metrics(session, metrics, options.host, options.port)
 
     status_ok()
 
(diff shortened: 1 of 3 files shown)


dankolbrs:
Output:
```
# python rabbitmq_status.py --username test --password test -n aio1-rabbit-mq-container-df66e8e1
status okay
metric rabbitmq_msgs_excl_notifications int64 0 messages
metric rabbitmq_sockets_total int64 3594 fd
metric rabbitmq_mem_used int64 120068080 bytes
metric rabbitmq_vhost_root_consumers int64 0 consumers
metric rabbitmq_fd_total int64 4096 fd
metric rabbitmq_notification_messages int64 0 messages
metric rabbitmq_uptime int64 264841570 ms
metric rabbitmq_deliver_get int64 252945 messages
metric rabbitmq_max_channels_per_conn int64 1 channels
metric rabbitmq_fd_used int64 160 fd
metric rabbitmq_publish int64 251508 messages
metric rabbitmq_messages_unacknowledged int64 0 messages
metric rabbitmq_vhost_heat_consumers int64 20 consumers
metric rabbitmq_mem_alarm_status uint32 1
metric rabbitmq_vhost_nova_consumers int64 25 consumers
metric rabbitmq_get int64 0 messages
metric rabbitmq_vhost_glance_consumers int64 0 consumers
metric rabbitmq_deliver int64 252945 messages
metric rabbitmq_sockets_used int64 108 fd
metric rabbitmq_vhost_keystone_consumers int64 0 consumers
metric rabbitmq_proc_total int64 1048576 processes
metric rabbitmq_ack int64 252945 messages
metric rabbitmq_proc_used int64 1783 processes
metric rabbitmq_total_consumers int64 142 consumers
metric rabbitmq_messages int64 0 messages
metric rabbitmq_disk_free_alarm_status uint32 1
metric rabbitmq_vhost_neutron_consumers int64 86 consumers
metric rabbitmq_vhost_cinder_consumers int64 11 consumers
metric rabbitmq_mem_limit int64 3347044761 bytes
metric rabbitmq_messages_ready int64 0 messages
```

dankolbrs:
Thanks @BjoernT . Updated to only report vhosts without consumers, added alarm criteria and kept current criteria flow by setting default of 0.

cloudnull:
This seems fine to me, however, I'm waiting on @BjoernT to approve.

hughsaunders:
This looks like its going to alert immediately as there are at least three vhosts with no consumers, is that a result of your testing environment and will be different in production?

dankolbrs:
@hughsaunders , two AIOs I was testing this on both were passing smoke tests and functioning correctly, but had no consumers on the vhosts `glance`, `root`, or `keystone`. Which is why I originally went the route of reporting individual vhosts. Per @BjoernT , all vhosts on production systems have consumers and this would report as expected. I'm unable to verify due to not having access to a production system.

BjoernT:
OK I realized the check is wrong. You have to look at each individual queues and list the consumers and report if there is a queue without consumers (accumulated per vhost). Don't list the consumers for a  vhost because keystone for example would "never" have queues to begin with

alextricity25:
> two AIOs I was testing this on both were passing smoke tests and functioning correctly, but had no consumers on the vhosts glance, root, or keystone.

Hmmmm...so no alarms were triggered even though you had 3 vhost without consumers?

dankolbrs:
@alextricity25 I was running the plugin manually, not via maas.

files/plugins/rabbitmq_status.py
```diff
@@ -172,6 +174,29 @@ def _get_queue_metrics(session, metrics, host, port):
...
+    for vhost in vhosts_response:
+        VHOSTS.append(vhost['name'])
+
+    queues_without_consumers = 0
+    for vhost in VHOSTS:
+        # uri encode /
+        URI_ENCODED = vhost.replace('/', '%2F')
+        queue_response = _get_rabbit_json(session, CONSUMERS_QUEUES_URL %
+                                          (host, port, URI_ENCODED))
+    for queue in queue_response:
```
git-harry: By the looks of it this for-loop needs indenting otherwise `queue_response` will only ever relate to the last value of `vhost`.

dankolbrs: Good call, thanks! Fixed.

dankolbrs:
recheck_cit_all - pip timeout

cloudnull:
recheck_cit_all

dankolbrs:
@BjoernT I've resolved all the issues from the review, anything further you'd like to see for an approval?

PR #242, opened 2017-05-05
Provide more flexible and reliable apt repo config

odyssey4me:
This patch adds the ability to use a different apt
repository for the maas packages using an alternative
key set and URL configuration. This is to facilitate
the ability to artifact the MaaS apt repo and
configure RPC-O to use the artifacted repo instead of
the upstream source.

Retries are added to the apt key addition to improve
reliability and the conditional for the repo/key
additions are removed as the values are defaulted
and are therefore always defined.

Connects https://github.com/rcbops/u-suk-dev/issues/1657

### DIFF
--- a/tasks/maas_host_install.yml
+++ b/tasks/maas_host_install.yml
@@ -15,12 +15,16 @@
 
 - name: Add MaaS apt keys
   apt_key:
-    url: "{{ item.url }}"
+    id: "{{ item.hash_id | default(omit) }}"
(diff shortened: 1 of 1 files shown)


evrardjp:
Let's wait for gates, but it should be cool.

odyssey4me:
OK, instead of rebasing on master (which is a little complicated right now), I'll cherry-pick the rpc-maas inception into this branch then adjust the rpc-maas role SHA once the other patch merges.

PR #245, opened 2017-05-05
Put back maas_rpc_scripts_dir variable

alextricity25:
This variable is used by the verify-maas.yml playbook in the RPCO
integrated repo.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -548,6 +548,10 @@ maas_pip_container_packages:
 
 maas_source_plugin_dir: plugins/
 
+# NOTE: (alextricity25) This variable is used by verify-maas.yml playbook
+# in the RPCO integrated repository.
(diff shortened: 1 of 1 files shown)


defaults/main.yml
```diff
@@ -548,6 +548,10 @@ maas_pip_container_packages:
 
 maas_source_plugin_dir: plugins/
 
+# NOTE: (alextricity25) This variable is used by verify-maas.yml playbook
+# in the RPCO integrated repository.
+maas_rpc_scripts_dir: /opt/rpc-openstack/scripts
```
cloudnull: This is not a variable used in this role so it has no reason to be here. IMO this should be renamed within the rpc-o repo as it is used there.

alextricity25:
I'm closing this off in favor of defining the variable in the verify-maas.yml playbook.

PR #246, opened 2017-05-05
Fixes MaaS for Ocata SHA

cjloader:
This fix updates the maas plugins to be able to be run with the upstream constraints
for novaclient and psutil.

Connects rcbops/u-suk-dev#1642

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -543,7 +543,7 @@ maas_pip_packages:
   - cryptography
   - ipaddr
   - lxml
-  - psutil<=1.2.1
+  - psutil<=5.0.1
   - apache-libcloud<2.0.0
   - rackspace-monitoring-cli
   - python-cinderclient
@@ -552,7 +552,7 @@ maas_pip_packages:
   - python-keystoneclient
   - python-magnumclient
   - python-neutronclient
-  - python-novaclient<7.0.0
+  - python-novaclient<=7.1.0
   - python-memcached
   - requests
   - swift
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -535,15 +535,11 @@ maas_requires_pip_packages:
 #
 # maas_pip_packages: These packages are installed inside the virtualenv.
 #
-# NOTE: python-novaclient 7.x changes how the list() method works and requires
-#       more work.
-# NOTE: psutil's API completely changed after 1.x, so we're keeping at 1.2.1
-#       until we can get it sorted out.
 maas_pip_packages:
   - cryptography
   - ipaddr
   - lxml
-  - psutil<=5.0.1
+  - psutil
   - apache-libcloud<2.0.0
   - rackspace-monitoring-cli
   - python-cinderclient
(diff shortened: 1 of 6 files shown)


alextricity25:
Squash and merge, please!

files/plugins/process_check_host.py
```diff
@@ -59,7 +61,13 @@ def get_processes(parent_pid=None):
         procs = [x for x in psutil.process_iter()
                  if x.pid not in excluded_pids]
     else:
-        procs = [x for x in psutil.Process(parent_pid).get_children()]
+        # psutil 1.2.1 has a children() method for the process object
+        # but the latest psutil uses get_children(). We need to handle
+        # this carefully.
```
evrardjp: I'm not really a good python dev, but shouldn't this be simply a try:  instead?

prometheanfire: Ya, that's the general way of handling it.  Could also look for the method using an if statement instead of falling back to try/except (which could catch other bad behavior where we don't want it to).

evrardjp: Yeah I won't block if not try: ...

alextricity25: I'll address that now.

alextricity25: done. Please look over the changes @evrardjp  and @prometheanfire

alextricity25:
Update: 
Ready for another look @evrardjp, @prometheanfire.

defaults/main.yml
```diff
@@ -552,7 +548,7 @@ maas_pip_packages:
   - python-keystoneclient
   - python-magnumclient
   - python-neutronclient
-  - python-novaclient<7.0.0
+  - python-novaclient
```
prometheanfire: do we care if this will pull in pike?

alextricity25: We should absolutely care! IMO we should put in some effort into converting these openstack service clients to use the openstacksdk. Similar to what is done here: https://github.com/rcbops/rpc-maas/pull/247

However, this isn't necessary for the issue-at-hand.

prometheanfire: well, ocata is 'unofficially' capped at <7.2.0 or the very least <8.0.0 (which we know is pike)

alextricity25: @prometheanfire I don't seem to understand. What is the suggestion/question here?

prometheanfire: if we care about using the same version as whatever branch we are on we should tie the pip install to upper-constraints.txt

alextricity25: Gotcha. Like what is done here? https://github.com/openstack/openstack-ansible-os_nova/blob/stable/ocata/tasks/nova_install.yml#L46-L51

alextricity25: Done. Please take another look @prometheanfire

prometheanfire: yep, just like that

alextricity25: @prometheanfire Awesome! Please give it an approved review if everything looks good to you.

PR #247, opened 2017-05-08
Replace glanceclient with openstacksdk

alextricity25:
This commit is the first of serveral in an effort to replace
all openstack service clients with the openstacksdk. This commit
does the following:

* Adds the get_openstack_conn() method to maas_common.py. This
  method returns back the Connection object required to perform
  operations on openstack resource objects.
* Modifies the glance_api_local_check plugin to use the openstacksdk
  instead of querying the glance endpoint directly.
* Adds the openstacksdk pip package to defaults/main.

### DIFF
--- a/files/plugins/glance_api_local_check.py
+++ b/files/plugins/glance_api_local_check.py
@@ -16,12 +16,9 @@
 
 import argparse
 import collections
-import ipaddr
 import time
 
-from glanceclient import exc as exc
-from maas_common import get_auth_ref
-from maas_common import get_glance_client
+from maas_common import get_openstack_conn
 from maas_common import metric
 from maas_common import metric_bool
 from maas_common import print_output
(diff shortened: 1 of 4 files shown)


files/plugins/glance_api_local_check.py
```diff
@@ -16,12 +16,9 @@
```
cloudnull: I'd rather see this effort put into the monitorstack plugin manager found here https://github.com/openstack/monitorstack than for us to continue to carry these plugins.

files/plugins/maas_common.py
```diff
@@ -87,55 +87,6 @@ def get_cinder_client(previous_tries=0):
```
cloudnull: See comment on files/plugins/glance_api_local_check.py

cloudnull:
@alextricity25 @cjloader Change looks good and IMO is going the right direction, however, I'd rather us not continue modifying/updating our old plugins. Instead, I'd rather us move the plugins upstream and remove the legacy plugins as we're migrating.

odyssey4me:
@alextricity25 Agree with @cloudnull here. Considering that this repo effort is targeting only Newton 14.1.0 onwards, I would suggest that we rather pile into the monitorstack effort for all the checks and leave this role to be focused purely on the implementation and configuration of MaaS to consume those plugins.

alextricity25:
Makes sense to me! Thanks for the feedback, gentlemen.

PR #256, opened 2017-05-26
Implement Role Breakdown

cloudnull:
This change implements the role break down approved in issue 255. This
change allows us to move away from the single role into a complete set
of playbooks which provides additional flexibility and speed when
deploying RPC-MaaS. This change also allows us to begin exploring ways
to add additional metric collection capabilities to the stack without
having to deal with the technical debt accumulated in the legacy role.

High-Level Overview:
  The RPC-MaaS monitoring system is comprised of a set of playbooks
allowing the deployer to either deploy everything using a single command
or to choose which components are required to monitor a stack. This
means that the monitoring components have been separated out from one
another allowing folks to deploy and monitor stacks like stand alone
swift, cinder, ceph, etc.

Upgrade-ability:
  Care has been taken to ensure that all of the checks deployed by this
change are backwards compatible with all previously known versions of an
RPC-MaaS deployment. This means deployers should be able to immediately
transition to this new deployment methodology without impacting existing
monitoring capabilities.

Usage:
  While these playbooks can be used with an OpenStack-Ansible deployment
they can also be used with a stand-alone inventory. See the
`test/inventory` file for more on how it can be used. As stated previously,
deployers are no longer required to have complete OpenStack ready
inventory, they can now pick and choose the components required without
having to worry about missing items and unknown handling of checks.

Testing:
  These playbooks have also been built to be 100% idempotent. At present
the testing is being done on a complete OpenStack cloud however in
future PRs we'll break down testing to individual components and ensure
that all of the playbooks are testable on their own.


cloudnull:
recheck_cit_all

evrardjp:
I'd be enclined to separate ansible 1.9 (using this method), and ansible>=2.0 into different branches.
This way we can improve readability and performance of the 2.0 using other mechanisms than a split out of a role, limiting re-use of the powerful ansible features.

evrardjp:
I will have a look at this to see if I can think of something outside the box that works for 1.9.

cloudnull:
closed and re-opened to un-block the gate.

cloudnull:
recheck_cit_all

cloudnull:
closed and re-opened to un-block the gate.

meta/main.yml
```diff
@@ -1,34 +0,0 @@
----
```
prometheanfire: Are we getting rid of this?

playbooks/files/pip-constraints.txt
```diff
@@ -0,0 +1,25 @@
+### Monitorstack plugin framework
```
prometheanfire: is there a reason not to use GR/UC for this type of thing?

playbooks/files/plugins/maas_common.py
```diff
@@ -34,7 +34,7 @@
                 'OS_IDENTITY_API_VERSION': None,
                 'OS_AUTH_VERSION': None,
                 'OS_ENDPOINT_TYPE': None,
-                'OS_API_INSECURE': False}
+                'OS_API_INSECURE': True}
```
prometheanfire: Is this settable?  I can imagine some customers actually run proper ssl certs.

playbooks/files/plugins/maas_common.py
```diff
@@ -43,7 +43,7 @@
 # v2 keystone auth is still necessary until RPCR switches over to v3 auth
 
 
-OPENRC = '/root/openrc-maas'
+OPENRC = '/root/openrc'
```
prometheanfire: This also should probably be settable, though it's able to be overridden if called correctly in the get_auth_info method.

playbooks/templates/network_throughput.yaml.j2
```diff
@@ -1,6 +1,6 @@
 {% set label = "network_throughput_"+item.0.name %}
 {% set check_name = label+'--'+ansible_hostname %}
-{% set max_speed = (item.0.max_speed | default(0, boolean=true) or item.1.stdout) | int * 10 ** 6 / 8 %}
```
prometheanfire: What's with the item1.stdout thing?  seems like a logic change to me at first glance.

meta/main.yml
cloudnull: Yes this is no longer a role so the meta has been removed.

playbooks/files/pip-constraints.txt
cloudnull: We eventually will be removing most of these when we move to ``monitorstack``. For now, we're maintaining our old existing constraints.

playbooks/files/plugins/maas_common.py
cloudnull: Yes if this is set in the openrc file it will use whatever is set there.

playbooks/files/plugins/maas_common.py
cloudnull: this is using the openrc file from osa instead of trying to carry our own version. This ensures we're not carrying useless code.

playbooks/templates/network_throughput.yaml.j2
cloudnull: We still have the option to specifically set the speed however the auto discovery portion of this is taken care of for us by the rax-maas agent.

prometheanfire:
you're welcome

Issue #264, opened 2017-06-05
stable/kilo: Update container_storage_checks.yaml.j2


npawelek:
The stable/kilo branch has a bug in the container_storage_checks template which deploys with a venv. As venv's don't yet exist in Kilo, it causes the check to fail.

https://github.com/rcbops/rpc-maas/blob/stable/kilo/templates/container_storage_checks.yaml.j2#L9-L10

Submitting a PR to update the template to use the plugin directly.

BjoernT:
That was addressed with https://github.com/rcbops/rpc-maas/commit/236a4082ad95e98ef94800b7ad3f0a0fef74612b did you use an older code base ?

PR #265, opened 2017-06-05
Remove maas_plugin_dir and fix associated templates

npawelek:
Update container_storage_checks.yaml.j2 to remove venv wrapper in stable/kilo.

Fixes: #264

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -485,8 +485,6 @@ maas_pip_dependencies:
 
 maas_source_plugin_dir: plugins/
 
-maas_plugin_dir: /usr/lib/rackspace-monitoring-agent/plugins/
-
 maas_rpc_scripts_dir: /opt/rpc-openstack/scripts
 
 # The following two variables control which checks and alarms are excluded from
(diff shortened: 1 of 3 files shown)


BjoernT:
👍  It seems to me, that maas_plugin_dir is now obsolete with this change, so let's remove it

npawelek:
Hrm, good catch. Looks like the only other place `maas_plugin_dir` is still used is within hp-check.yaml.j2[1]. Let me update that template as it's broken too, then I'll remove the `maas_plugin_dir` variable entirely.

[1] https://github.com/rcbops/rpc-maas/blob/stable/kilo/templates/hp-check.yaml.j2#L9-L10

npawelek:
Updated. How's this look @BjoernT?

cloudnull:
@npawelek @BjoernT is the removal of this being done as a kilo specific change? Something that was accidentally backported?

npawelek:
@cloudnull I'm not sure why the stable/kilo and stable/liberty branches have references to venv's in the plugin templates. They fail upon install, as the venv wrapper wasn't included until Mitaka. I'll need to replicate these same changes to stable/liberty.

cloudnull:
sounds good to me !

PR #271, opened 2017-06-06
Build in Ansible 1.9 support

cloudnull:
These changes ensure that the playbooks used to deploy maas are both 2.x
and 1.9 compatible at the same time.

This change also removes the OSA library and role dependencies which are
platform specific.

### DIFF
--- a/playbooks/maas-infra-rabbitmq.yml
+++ b/playbooks/maas-infra-rabbitmq.yml
@@ -16,8 +16,24 @@
 - name: Gather facts
   hosts: rabbitmq_all
   gather_facts: "{{ gather_facts | default(true) }}"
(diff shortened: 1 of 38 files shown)


playbooks/maas-infra-rabbitmq.yml
```diff
@@ -16,10 +16,26 @@
...
+      until: install_pip_packages|success
+      retries: 5
+      delay: 2
+
+  vars_files:
+    - vars/main.yml
+    - vars/maas-infra.yml
   tags:
-    - maas-infra-rabbitmq
+    - maas-infra-memcached
```
cloudnull: good catch. fixed.

PR #280, opened 2017-06-08
Excluding empty queues from consumer check

BjoernT:
Queues having no consumers subscribed and don't
contain any messages are excluded from the
consumer check.
This is done to reduce alerts on non critical conditions.


cloudnull:
Just waiting on the gate to complete before merging.

PR #281, opened 2017-06-09
Thread the telegraf entrypoint

cloudnull:
Running all plugins from MaaS in a single pass is slow, this chnage
implements threading allowing the collection of the plugins to happen
much faster.

Three bugs were discovered:
  * Should a container be stopped the pid returned for the container is
    "-1" which will fail a lookup, a change has been added to condition has
    been added to check for a negative pid and skip when needed.
  * The horizon dashboard variable had a space in it causing the lookup
    to fail. This removes the extra bits and corrects the plugin
    failure.
  * The venv wrapper used to source the virtual env and then run the a
    python utility while this mostly works it breaks a shell environment
    should the check fail to run. Calling the virtual environment python
    directly resolves this issue.

### DIFF
--- a/playbooks/files/maas_telegraf_format.py
+++ b/playbooks/files/maas_telegraf_format.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import multiprocessing
 import os
 import subprocess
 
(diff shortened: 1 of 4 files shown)


evrardjp:
The third bug is what I saw in the past. Good fix, let me review that quickly :)

playbooks/files/maas_telegraf_format.py
```diff
@@ -22,6 +23,52 @@
...
+        while True:
+            detail_args = self.queue.get()
+            if not detail_args:
+                break
+
+            detail_args.insert(0, RUN_VENV)
+            detail_args.insert(2, '--telegraf-output')
+            with open(os.devnull, 'w') as null:
+                process = subprocess.Popen(
+                    ' '.join(detail_args),
```
evrardjp: Why do we do this, instead of calling directly the appropriate python bits of each plugin? 
Is there a plan to change to that ?

cloudnull: This is happening because each plugin is executed out of a venv which would require changing the interpreter (activating the venv), then running each plugin within that interpreter, which I'd rather not have to deal with. In the future, this will all be replaced with monitorstack so it will all eventually be deleted.

PR #282, opened 2017-06-11
Added telegraf and influx metric collection

cloudnull:
Two new playbooks have been added to deploy telegraf and influxdb into
an environmeent and begin collecting metrics locally using the same
plugins powering the rax-maas check system.

In the course of adding this collection capability I discovered the
plugin `process_check_container.py` was throwing an error should a given
container lookup fail. To correct this I've added an exception handler.

### DIFF
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -83,20 +83,6 @@ In order to enable console auth checking set the ``nova_console_type`` and
     echo 'nova_console_port: 6082' | tee -a /etc/openstack_deploy/user_variables.yml
 
 
-Collecting metrics for time series analysis
--------------------------------------------
-
-An available plugin can be executed at any time to run all discovered checks
-found in ``/etc/rackspace-monitoring-agent.conf.d``. These plugins will be
-executed having the output format converted to telegraf line format. This is
-useful when storing local checks in a time series database like influxdb. To
-run the plugin execute the following:
-
-.. code-block:: bash
-
-    /openstack/venvs/maas-${VERSION_NUM}/bin/python /usr/lib/rackspace-monitoring-agent/plugins/maas_telegraf_format.py
-
-
 Recommended Overrides for specific versions of OpenStack-Ansible
 ----------------------------------------------------------------
 
@@ -248,6 +234,26 @@ variable required to be set is ``maas_swift_accesscheck_password``.
     ansible-playbook -e @/etc/openstack_deploy/user_secrets.yml /opt/rpc-maas/playbooks/maas-openstack-swift.yml -i inventory
 
 
+maas-tigkstack-influxdb.yml
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``maas-tigkstack-influxdb.yml`` playbook will create an influxdb user.
+For this to happen a password must be set in a secrets file somewhere and
+referenced during the playbook run. The variables required are
+``influxdb_db_root_password`` and ``influxdb_db_metric_password``.
+
+.. code-block:: bash
+
+    # Variable file set
+    echo 'influxdb_db_root_password: secrete' | tee -a /etc/openstack_deploy/user_secrets.yml
+    echo 'influxdb_db_metric_password: secrete'  | tee -a /etc/openstack_deploy/user_secrets.yml
+
+    # Example playbook run with variable file.
+    ansible-playbook -e @/etc/openstack_deploy/user_secrets.yml /opt/rpc-maas/playbooks/maas-tigkstack-influxdb.yml -i inventory
+
+
+
+
 Restart flow control
 --------------------
 
@@ -260,3 +266,50 @@ redeploying specific checks, this variable is not needed.
 .. code-block:: bash
 
     ansible-playbook -e "maas_restart_independent=false" /opt/rpc-maas/playbooks/site.yml
+
+
+Deploying the TIGK Stack
+------------------------
+
+If ``the maas-tigkstack-all.yml`` Playbook is run **telegraf** will be
+installed throughout the deployment and **influxdb** will be installed
+within the *log_host*. Enabling this service will immediately being
+collecting metrics within the deployment using the existing MaaS plugins
+and enabled checks.
+
+
+Collecting metrics for time series analysis
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+An available plugin can be executed at any time to run all discovered checks
+found in ``/etc/rackspace-monitoring-agent.conf.d``. The existing MaaS plugins
+will be executed having the output format converted to telegraf line format.
+This is useful when storing local checks in a time series database like
+influxdb and is part of the normal TIGK Stack deployment and **telegraf**
+agent configuration.
+
+
+To run the plugin execute the following:
+
+.. code-block:: bash
+
+    /openstack/venvs/maas-${VERSION_NUM}/bin/python /usr/lib/rackspace-monitoring-agent/plugins/maas_telegraf_format.py
+
+
+Setting the job reference
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When deploying the TIGK Stack you can set the job reference variable to create
+a tag on all metrics collected for the specific job. The variable setting is
+``maas_job_reference`` which has a default of **testing**. Setting this job
+variable will provide the ability to select and search through aggregated
+metrics using a simple where clause, example:
+
+.. code-block:: sql
+
+    use telegraf
+    select * from nova_api_metadata where job_reference='testing'
+
+
+This simple query will select everything from the **nova_api_metadata** metric
+where the job reference tag is equal to the string "testing".
--- a/playbooks/files/plugins/container_storage_check.py
+++ b/playbooks/files/plugins/container_storage_check.py
@@ -87,7 +87,6 @@ def main():
         status_err(str(e))
     else:
         status_ok()
-    finally:
         metric_bool(
             'container_storage_percent_used_critical',
             _container_check
--- a/playbooks/files/tigkstack/deploy_go.sh
+++ b/playbooks/files/tigkstack/deploy_go.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+
+pushd /opt
+  wget -P /opt/ https://storage.googleapis.com/golang/go1.8.3.linux-amd64.tar.gz
+  tar  -xzf /opt/go1.8.3.linux-amd64.tar.gz -C /opt/
+popd
+
+pushd /usr/local/bin
+  find /opt/go/bin/ -type f -exec ln -sf {} \;
+popd
+
+if ! grep -qw 'GOROOT="/opt/go"' /etc/environment; then
+  echo 'GOROOT="/opt/go"' | tee -a /etc/environment
+fi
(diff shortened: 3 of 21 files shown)


doc/source/usage.rst
```diff
@@ -260,3 +266,50 @@ redeploying specific checks, this variable is not needed.
...
     ansible-playbook -e "maas_restart_independent=false" /opt/rpc-maas/playbooks/site.yml
+
+
+Deploying the TIGK Stack
+------------------------
+
+If ``the maas-tigkstack-all.yml`` Playbook is run **telegraf** will be
+installed throughout the deployment and **influxdb** will be installed
+within the *log_host*. Enabling this service will immediately being
+collecting metricss within the deployment using the existing MaaS plugins
```
alextricity25: s/metricss/metrics/

cloudnull: done.

playbooks/files/tigkstack/deploy_go.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+pushd /opt
+  wget -P /opt/ https://storage.googleapis.com/golang/go1.7.3.linux-amd64.tar.gz
```
prometheanfire: should this be a variable or a better url?

playbooks/templates/tigkstack/influxdb.conf.j2
```diff
@@ -0,0 +1,80 @@
...
+  pprof-enabled = false
+  lease-duration = "1m0s"
+
+[data]
+  enabled = true
+  dir = "/var/lib/influxdb/data"
+  wal-dir = "/var/lib/influxdb/wal"
+  wal-logging-enabled = true
+  query-log-enabled = false
+  cache-max-memory-size = 679477248
```
prometheanfire: should some of these be calculated values?

tests/rpc-maas-overrides.yml
```diff
@@ -125,3 +125,10 @@ neutron_venv_tag: "testing"
 nova_venv_tag: "testing"
 sahara_venv_tag: "testing"
 swift_venv_tag: "testing"
+
+
+# Set vars for tigk stack
+influxdb_db_root_password: SuperSecrete
+influxdb_db_metric_password: SuperDuperSecrete
```
prometheanfire: shouldn't these passwords be switched (root and metric)

cloudnull: just used in tests. but i can change if you think it best.

playbooks/files/tigkstack/deploy_go.sh
cloudnull: Updated to latest 1.8 however it doesn't look like go provides a "latest" link. If we can find one then I'd be happy to switch it up.

playbooks/templates/tigkstack/influxdb.conf.j2
cloudnull: At this point no.

cloudnull: At this point no.

cloudnull: At this point no.

prometheanfire: k

playbooks/maas-tigkstack-influxdb.yml
```diff
@@ -0,0 +1,114 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Deploy influxdb
+  hosts: log_hosts
+  gather_facts: true
+  pre_tasks:
+    - name: Check init system
```
evrardjp: Please use ansible fact ``{{ ansible_service_mgr }}`` instead of building your own.

playbooks/maas-tigkstack-influxdb.yml
```diff
@@ -0,0 +1,114 @@
...
+    - name: Add influxdata apt-keys
+      apt_key:
+        url: "https://repos.influxdata.com/influxdb.key"
+        state: "present"
+
+    - name: Add influxdata repo
+      apt_repository:
+        repo: "deb https://repos.influxdata.com/{{ ansible_distribution | lower }} {{ ansible_distribution_release }} stable"
+        state: "present"
+        update_cache: yes
```
evrardjp: nit:
``update_cache: yes`` is default for apt_repository, and should work fine. No need to explicitly say it.

playbooks/maas-tigkstack-influxdb.yml
```diff
@@ -0,0 +1,114 @@
...
+        - always
+
+    - name: Set the name of pid1
+      set_fact:
+        pid1_name: "{{ _pid1_name.stdout }}"
+      tags:
+        - always
+
+    - name: Add influxdata apt-keys
+      apt_key:
```
evrardjp: If we intend users to deploy influx, maybe we should make this task (and the next one) conditional, to make sure they use what would be in our repos.

playbooks/maas-tigkstack-influxdb.yml
```diff
@@ -0,0 +1,114 @@
...
+      template:
+        src: templates/tigkstack/relay.toml.j2
+        dest: /opt/influxdb-relay/relay.toml
+
+    - name: Drop Influx Relay upstart
+      template:
+        src: templates/tigkstack/influxdbrelay.conf.j2
+        dest: /etc/init/influxdbrelay.conf
+      when:
+        - pid1_name == "init"
```
evrardjp: See comment above with service mgr

playbooks/maas-tigkstack-telegraf.yml
```diff
@@ -0,0 +1,80 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Deploy telegraf
+  hosts: "hosts:all_containers"
+  gather_facts: true
+  tasks:
+    - name: Add influxdata apt-keys
+      apt_key:
+        url: "https://repos.influxdata.com/influxdb.key"
```
evrardjp: same remark as above to make it conditional.

playbooks/maas-tigkstack-telegraf.yml
```diff
@@ -0,0 +1,80 @@
...
+        src: templates/tigkstack/telegraf.conf.j2
+        dest: /etc/telegraf/telegraf.conf
+      register: telegraf_config
+
+    - name: Set telegraf defaults
+      copy:
+        content: |
+          USER="root"
+          GROUP="root"
+        dest: /etc/default/telegraf
```
evrardjp: would that work with systemd, now that we're thinking ocata too?

playbooks/templates/tigkstack/grafana.ini.j2
```diff
@@ -0,0 +1,65 @@
...
+[log.syslog]
+
+[event_publisher]
+
+[dashboards.json]
+
+[metrics]
+
+[grafana_net]
+url = https://grafana.net
```
evrardjp: Probably we should shamelessly put an url to something else ? I don't know, I know nothing about that. I think other reviewers welcomed on this.

playbooks/templates/tigkstack/influxdb.conf.j2
evrardjp: At this point no? :p

playbooks/templates/tigkstack/telegraf.conf.j2
```diff
@@ -39,6 +39,12 @@
   write_consistency = "any"
   timeout = "30s"
 
+{% if outputs_prometheus_client is defined %}
+[[outputs.prometheus_client]]
+  listen = "{{ outputs_prometheus_client_listen|default(':9126') }}"
```
cloudnull: This should be defined in the maas-tigkstack.yml variable file.

cloudnull: the variable should also follow the same prefix ``telegraf_outputs_prometheus_client_listen``.

playbooks/templates/tigkstack/telegraf.conf.j2
```diff
@@ -39,6 +39,12 @@
...
+  expiration_interval = "{{ outputs_prometheus_client_expiration_interval|default('60s') }}"
```
cloudnull: same as above

playbooks/vars/maas-tigkstack.yml
```diff
@@ -40,3 +40,8 @@ influxdb_relay_port: 9096
 influxdb_protocol: http
 
 influx_telegraf_targets: []
+
+# Prometheus Client Outputs Plugin
+#outputs_prometheus_client: true
```
cloudnull: These should be under the telegraf_.* variable type.

playbooks/files/plugins/container_storage_check.py
```diff
@@ -87,7 +87,6 @@ def main():
         status_err(str(e))
     else:
         status_ok()
-    finally:
```
cloudnull: It looks like this raises an exception https://github.com/rcbops/rpc-maas/blob/master/playbooks/files/plugins/maas_common.py#L563-L571 which would produce a stacktrace making it impossible for maas to record the metrics even if something was produced.

playbooks/maas-tigkstack-influxdb.yml
cloudnull: done.

playbooks/maas-tigkstack-influxdb.yml
cloudnull: ++ I think we can lock it all down to a specific repo in a follow on patch.

playbooks/maas-tigkstack-influxdb.yml
cloudnull: This is not available in Ansible 1.9.

playbooks/maas-tigkstack-influxdb.yml
cloudnull: This is not available in ansible 1.9

playbooks/maas-tigkstack-telegraf.yml
cloudnull: this can be done in the next patch set.

playbooks/maas-tigkstack-telegraf.yml
cloudnull: yes. based on what I've been able to ascertain this works, however, I need to build some tests to validate.

playbooks/maas-tigkstack-influxdb.yml
evrardjp: Ok I'll create a card for that.

doc/source/usage.rst
```diff
@@ -248,6 +234,26 @@ variable required to be set is ``maas_swift_accesscheck_password``.
...
+The ``maas-tigkstack-influxdb.yml`` playbook will create an influxdb user.
+For this to happen a password must be set in a secrets file somewhere and
+referenced during the playbook run. The variables required are
+``influxdb_db_root_password`` and ``influxdb_db_metric_password``.
+
+.. code-block:: bash
+
+    # Variable file set
+    echo 'influxdb_db_root_password: secrete' | tee -a /etc/openstack_deploy/user_secrets.yml
+    echo 'influxdb_db_metric_password: secrete'  | tee -a /etc/openstack_deploy/user_variables.yml
```
cloudnull: done.

Issue #286, opened 2017-06-19
Update MaaS RabbitMQ check to exclude versioned_notification queues


BjoernT:
https://github.com/rcbops/rpc-maas/blob/master/playbooks/templates/rax-maas/plugins/rabbitmq_status.py#L170-171
    
The old method of filtering based on .startswith won't catch the versioned_notifications.X (info,error,etc.) queues, causing false positive alerts when those queues build up.

This seems to only be an issue starting with the Mitaka release (when versioned notifications were introduced), and then only if the OpenStack services are configured to use the notification driver messagingv2. If not configured via the notification_format parameter, the default is to send both versioned and unversioned notifications.

## What were you doing when it happened?

This issue was observed with RPC-R which uses messagingv2 as the notification driver defaults 

I am unsure when/if the notification driver has been changed from messaging to messagingv2 in OSA, but assumed a pro-active approach was better.  We have updated our fork of the checks to include this change.

BjoernT:
The larger problem I see is to exclude those queues but not checking for consumers, ultimately leading into the situation where the notifications queues are always growing to a point where it can crash rabbitmq. The check would need to be dependent on the notification_driver configuration and if ceilometer is enabled or not. If you setup a notification driver other than noop but not configuring ceilometer, you setting yourself up for some headache and disabling monitoring would be a bad thing.

cloudnull:
Looks like this is now taken care of, can we close this issue?

PR #287, opened 2017-06-19
Exclude notifications queues under condition

BjoernT:
The notifications queues can only be execluded from monitoring
if one or more consumer is present on the queue.
If the queues are excluded from monitoring but no consumer is present,
it will lead to instabilities of RabbitMQ, ultimately stopping
operations due to system limitations (memory/disk).
The notification_driver need to be properly set to noop if the
messages generation for the notification queues should be suppressed.
Additionally the new versioned_notifications queue is included in this
check.

Closes-Bug: #286


cloudnull:
recheck_cit_all

cloudnull:
recheck_cit_rpc-maas_master_trusty_mitaka

PR #289, opened 2017-06-26
Drop keyring file on osds

npawelek:
Add task to drop the ceph keyring on the osds nodes as it's needed by the ceph_osd_stats check.

### DIFF
--- a/playbooks/maas-ceph-mon.yml
+++ b/playbooks/maas-ceph-mon.yml
@@ -25,17 +25,7 @@
   tags:
     - maas-ceph-mons
 
-- name: Install checks for ceph mons
-  hosts: mons[0]
-  gather_facts: false
-  pre_tasks:
-    - name: Create ceph monitoring client
-      command: ceph auth get-or-create client.raxmon mon 'allow r'
-      register: ceph_raxmon_client
-      tags:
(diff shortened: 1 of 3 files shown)


cloudnull:
This looks good however I think we need to add the playbook to capture this key here too. Maybe make it an include playbook? This bit could be an included playbook in both the osd and mons playbooks: https://github.com/npawelek/rpc-maas/blob/cb768bd769f92f60fb116e64b950e037437e08ba/playbooks/maas-ceph-mon.yml#L28-L38 making it so neither playbook is dependent on one

cloudnull:
I added to your commit.

npawelek:
It looks like this is properly dropping the raxmon keyring to all ceph nodes, however, the restart handler isn't functioning properly. It looks like both mons hosts are restarting and only the first of the osds.

npawelek:
I reverted these changes and I'm still seeing the issue. I believe the restart handler problem is a result of the current standalone inventory being used on this AIO.  We should be safe to merge these changes.

npawelek:
recheck_cit_rpc-maas_master_trusty_liberty
recheck_cit_rpc-maas_master_trusty_mitaka

PR #294, opened 2017-07-10
[LA-314] change log_hosts to influx_hosts

cloudnull:
This change creates a specific group for influxdb deployments. This
change is being done so that we dont assume the log hosts will be used
for metric collection.

### DIFF
--- a/playbooks/templates/tigkstack/relay.toml.j2
+++ b/playbooks/templates/tigkstack/relay.toml.j2
@@ -3,8 +3,8 @@ name = "example-http"
 bind-addr = '0.0.0.0:{{ influxdb_relay_port  }}'
 output = [
 {% set i =1%}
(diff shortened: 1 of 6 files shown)


playbooks/templates/tigkstack/relay.toml.j2
```diff
@@ -3,7 +3,7 @@ name = "example-http"
 bind-addr = '0.0.0.0:{{ influxdb_relay_port  }}'
 output = [
 {% set i =1%}
-{%for host_name in groups['log_hosts'] %}
+{%for host_name in groups['influx_hosts'] | default([]) %}
```
evrardjp: nit: whitespace before for:

cloudnull: done.

PR #297, opened 2017-07-12
Override raxmon ssl verification for non-transparent web proxy

npawelek:
The hardcoded nature of raxmon SSL verification can be problematic if a customer is using a non-transparent web proxy, which will fail verification checks. This allows an override for verification allowing the playbooks to properly set an entities agentID.

### DIFF
--- a/playbooks/library/raxmon.py
+++ b/playbooks/library/raxmon.py
@@ -164,9 +164,6 @@ def main():
 
     # Provide the ability to enable or disable SSL certificate verification
     # with raxmon
-    CA_CERT_PATH = os.path.join(os.path.abspath(os.path.dirname(__file__)),
-                           'data/cacert.pem')
-    libcloud.security.CA_CERTS_PATH.append(CA_CERT_PATH)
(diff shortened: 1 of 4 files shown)


playbooks/library/raxmon.py
```diff
@@ -160,6 +160,21 @@ def main():
     # importing the correct libraries
     from rackspace_monitoring.providers import get_driver
     from rackspace_monitoring.types import Provider
+    import libcloud.security
+
+    # Provide the ability to enable or disable SSL certificate verification
+    # with raxmon
+    CA_CERT_PATH = os.path.join(os.path.abspath(os.path.dirname(__file__)),
```
cloudnull: This is the __file__ path of the raxmon module. Are we storing, or do we expect to store a cacert here?

npawelek: Thanks. I confirmed, these are not needed and will be updating the PR shortly.

npawelek:
I think I did this right... How's it look now?

PR #299, opened 2017-07-12
UG-634 Add vars file for verification dependencies

mattt416:
Deploying rpc-maas on a multi-node fails on the verification stage,
because we're not explicitly installing the futures module on all
hosts.  It appears to be getting pulled in implicitly on some hosts
via other modules and therefore isn't failing across the board.

This commit simply creates a new vars file for the verification stage,
and updates maas-verify.yml to install this new pip package set.

Lastly, we remove futures from test-requirements.txt since it is now
installed via the maas-verify.yml playbook.

### DIFF
--- a/playbooks/maas-verify.yml
+++ b/playbooks/maas-verify.yml
@@ -46,6 +46,25 @@
         - rpc-maas-tool.py
         - alarmparser.py
 
+    - name: Copy over pip constraints
+      copy:
+        src: "files/pip-constraints.txt"
+        dest: "/tmp/pip-constraints.txt"
+
+    - name: Install verify pip packages to venv
+      pip:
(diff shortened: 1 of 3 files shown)


mattt416:
Please don't merge just yet -- still doing some testing here!

mattt416:
Tested, works for me.

cloudnull:
recheck_cit_rpc-maas_master_trusty_kilo

cloudnull:
recheck_cit_rpc_maas_master_trusty_kilo
--

cloudnull:
recheck_cit_rpc_maas_master_trusty_kilo

cloudnull:
recheck_cit_trusty_kilo

PR #300, opened 2017-07-13
use known measurement names

cloudnull:
Measurement names were random and hard to track, this change ensures
that all maas plugins using a known measurement name. This makes it easy
to write new queries and track changes without having to first figure
out the measurement name based on an inconsistent maas check label.


cloudnull:
recheck_cit_all

PR #302, opened 2017-07-13
Add additional constraints and vendor URI

cloudnull:
The constraints needed in the test repo need to be updated to ensure
pyrax works 100% of the time.

The URI module in early versions of Ansible is broken. To fix this the
URI module has been vendored to ensure that it will always work for our
limited needs. To allow this module to work its using the deprecated
syntax such that it will not conflict with the actual URI module as
provided by Ansible.


cloudnull:
recheck_cit_all

cloudnull:
recheck_cit_all

cloudnull:
recheck_cit_all

cloudnull:
recheck_cit_all

Issue #311, opened 2017-07-18
Split maas_external_ip_address into multiple vars


cfarquhar:
The `maas_external_ip_address` variable is currently used for both `target_hostname` and `url` when defining a remote HTTP MaaS check \[1].  Some use cases require differentiating between the two.  

As an example, my use case includes internal-only domain names and a public IP NAT'd to the internal API endpoint.  Thus, I need different values for `target_hostname` (i.e. public IP address 123.123.123.123) and the domain component of the `url` (i.e. openstack04.internal-only.domain).

\[1] Example remote HTTP MaaS check definition
```
{% set label = "lb_api_check_nova" %}
{% set check_name = label+'--'+maas_lb_name %}
type              : remote.http
label             : "{{ check_name }}"
period            : "{{ maas_check_period_override[label] | default(maas_check_period) }}"
timeout           : "{{ maas_check_timeout_override[label] | default(maas_check_timeout) }}"
disabled          : "{{ (inventory_hostname != groups['nova_api_os_compute'][0] or check_name | match(maas_excluded_checks_regex)) | ternary('true', 'false') }}"
target_resolver   : "IPv4"
target_hostname   : "{{ maas_external_ip_address }}"
details           :
    url           : "{{ maas_nova_scheme | default(maas_scheme)}}://{{ maas_external_ip_address }}:8774"
monitoring_zones_poll:
{% for zone in maas_monitoring_zones %}
  - {{ zone }}
{% endfor %}
```

PR #315, opened 2017-07-19
Add maas_external_hostname variable

cfarquhar:
This commit splits the `maas_external_ip_address` variable so that
different values can be configured in MaaS's `target_hostname` and
`url` settings.  This is required for environments where OpenStack
endpoints are not resolvable by the MaaS pollers, which requires
setting an IP for `target_hostname` and the internally resolvable
hostname in the `url`.

Closes: #311


cloudnull:
1 merge conflict due to other things going in today, I've rebased this change and will try and get it in later today.

PR #325, opened 2017-07-28
LA-378 Fix swift-recon release identifier

git-harry:
This change modifies the swift-recon MaaS plugin to allow the directory
containing swift-recon to be specified. This change removes the need for
the plugin to try to discover the version.

Checking `/etc/openstack-release` for the version to use in general
works however due to the ordering of playbooks in leapfrog upgrades this
causes the wrong version to be used. The use of this approach also
lengthens the time of plugin failure during an upgrade given the
separation between creating the release file and updating Swift.

### DIFF
--- a/playbooks/files/rax-maas/plugins/swift-recon.py
+++ b/playbooks/files/rax-maas/plugins/swift-recon.py
@@ -23,6 +23,7 @@
 # python swift-recon.py quarantine
 
 import argparse
+import os
 import re
 import shlex
 import subprocess
@@ -39,7 +40,7 @@ class CommandNotRecognized(maas_common.MaaSException):
     pass
 
 
-def recon_output(for_ring, options=None):
+def recon_output(for_ring, options=None, swift_recon_path=None):
     """Run swift-recon and filter out extraneous printed lines.
 
     ::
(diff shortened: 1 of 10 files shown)


git-harry:
recheck_cit_trusty_liberty
recheck_cit_trusty_mitaka
recheck_cit_xenial_newton
recheck_cit_xenial_ocata

playbooks/files/rax-maas/plugins/swift-recon.py
```diff
@@ -66,14 +66,14 @@ def recon_output(for_ring, options=None):
     """
 
     # grab the current release
-    with open("/etc/openstack-release") as f:
+    with open("/etc/rpc-release") as f:
```
cloudnull: Maybe we convert this to a loop and search for `rpc-release` and then `osa-release`  which would ensure the check works for both rpc-o and osa.

evrardjp:
@git-harry could you please update this?

Also, I have this: https://review.openstack.org/#/c/489531/ pending, I will test it, and it should improve the situation for you: When rpc maas will re-deploy, it will redeploy with the proper version.
But I still think this improvement above could be done too.

PR #326, opened 2017-07-28
Allow telegraph to schedule the inputs

evrardjp:
If we define multiple inputs, instead of multiple commands, the
inputs will be handled separately, instead of having a spike
of all the commands to run.

This lightens the load and allow more aggressive data collection
without ooming on low end machines.

### DIFF
--- a/playbooks/templates/tigkstack/telegraf.conf.j2
+++ b/playbooks/templates/tigkstack/telegraf.conf.j2
@@ -9,16 +9,24 @@
 {%   set telegraf_commands = [] %}
 {%   set _ = telegraf_commands.extend(telegraf_maas_commands) %}
 
-{% set command_interval = telegraf_commands | length * 3 | round %}
-{% set interval = ((telegraf_interval | int) > command_interval) | ternary(telegraf_interval, command_interval) %}
+{% set interval = (telegraf_interval | int) | round %}
+{#
+Each collection can take input_command_timeout seconds, and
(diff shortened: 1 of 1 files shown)


playbooks/templates/tigkstack/telegraf.conf.j2
```diff
@@ -13,12 +13,12 @@
 {% set interval = ((telegraf_interval | int) > command_interval) | ternary(telegraf_interval, command_interval) %}
 
 [agent]
-  interval = "{{ interval | round }}s"
+  interval = "20s"
```
cloudnull: This interval needs to be longer or variable.

playbooks/templates/tigkstack/telegraf.conf.j2
```diff
@@ -13,12 +13,12 @@
...
   round_interval = false
-  metric_batch_size = 1024
-  metric_buffer_limit = 10240
+  metric_batch_size = 2048
+  metric_buffer_limit = 20480
   collection_jitter = "8s"
-  flush_interval = "{{ interval + 10 | round }}s"
+  flush_interval = "30s"
```
cloudnull: This interval needs to be longer or variable.

cloudnull:
recheck_cit_xenial_ocata

evrardjp:
Now more variable. Evaluating this on my own machine.

PR #328, opened 2017-08-02
Turtles 106 ironic capacity check

tonytan4ever:
Review

### DIFF
--- a/playbooks/files/rax-maas/plugins/README.md
+++ b/playbooks/files/rax-maas/plugins/README.md
@@ -164,15 +164,16 @@ IP address of service to test
 #### ironic_capacity_check.py
 
 ##### Description:
-Polls a (native) ironic API living on the specified IP. Checks the percentage of nodes in maintenance--
+Polls a (native) ironic API living on the specified IP. Gather the number of total nodes and up nodes (nodes not in maintenance)
 
 ##### Mandatory Arguments:
 IP address of service to test
 ##### Example Output:
 
     status okay
-	metric client_success uint32 1
-	metric ironic_capacity double 0.00 %
+    metric client_success uint32 1
+    metric ironic_total_nodes_count uint32 1
+    metric ironic_up_nodes_count uint32 1
 
 ***
 #### magnum_api_local_check.py
--- a/playbooks/files/rax-maas/plugins/README.md
+++ b/playbooks/files/rax-maas/plugins/README.md
@@ -172,8 +172,9 @@ IP address of service to test
 
     status okay
     metric client_success uint32 1
-    metric ironic_total_nodes_count uint32 1
+    metric ironic_maint_nodes_count uint32 0
     metric ironic_up_nodes_count uint32 1
+    metric ironic_total_nodes_count uint32 1
 
 ***
 #### magnum_api_local_check.py
--- a/playbooks/files/rax-maas/plugins/README.md
+++ b/playbooks/files/rax-maas/plugins/README.md
@@ -161,6 +161,20 @@ IP address of service to test
     metric ironic_api_local_response_time double 42.690 ms
 
 ***
+#### ironic_capacity_check.py
+
+##### Description:
+Polls a (native) ironic API living on the specified IP. Checks the percentage of nodes in maintenance--
+
+##### Mandatory Arguments:
+IP address of service to test
+##### Example Output:
+
+    status okay
+	metric client_success uint32 1
+	metric ironic_capacity double 0.00 %
+
+***
 #### magnum_api_local_check.py
 
 ##### Description:
--- a/playbooks/files/rax-maas/plugins/README.md
+++ b/playbooks/files/rax-maas/plugins/README.md
@@ -164,7 +164,7 @@ IP address of service to test
 #### ironic_capacity_check.py
 
 ##### Description:
-Polls a (native) ironic API living on the specified IP. Gather the number of total nodes and up nodes (nodes not in maintenance)
+Polls a (native) ironic API living on the specified IP. Gathers the number of total nodes and up nodes (nodes not in maintenance)
 
 ##### Mandatory Arguments:
 IP address of service to test
@@ -172,7 +172,6 @@ IP address of service to test
 
     status okay
     metric client_success uint32 1
-    metric ironic_maint_nodes_count uint32 0
     metric ironic_up_nodes_count uint32 1
     metric ironic_total_nodes_count uint32 1
 
--- a/playbooks/files/rax-maas/plugins/ironic_capacity_check.py
+++ b/playbooks/files/rax-maas/plugins/ironic_capacity_check.py
@@ -15,7 +15,6 @@
 # limitations under the License.
 
 import argparse
-import time
 
 import ipaddr
 from ironicclient import exc
@@ -46,21 +45,21 @@ def check(auth_ref, args):
     except Exception as e:
         metric_bool('client_success', False, m_name='maas_ironic')
         status_err(str(e), m_name='maas_ironic')
+        return
     else:
         metric_bool('client_success', True, m_name='maas_ironic')
         # pass limit=0 to list all nodes list without pagination
         all_nodes = ironic.node.list(limit=0)
+        status_ok(m_name='maas_ironic')
 
-    status_ok(m_name='maas_ironic')
     if is_up:
         maint_nodes = [node for node in all_nodes if node.maintenance]
-        maint_nodes_percentage = len(maint_nodes)/len(all_nodes)
+        total_nodes = len(all_nodes)
+        up_nodes = total_nodes - len(maint_nodes)
         # question: is capacity the maint_nodes_percentage
         # or 1 - maint_nodes_percentage ?
-        metric('ironic_capacity',
-               'double',
-               '%.2f' % maint_nodes_percentage,
-               '%')
+        metric('ironic_total_nodes_count', 'uint32', total_nodes)
+        metric('ironic_up_nodes_count', 'uint32', up_nodes)
 
 
 def main(args):
--- a/playbooks/files/rax-maas/plugins/ironic_capacity_check.py
+++ b/playbooks/files/rax-maas/plugins/ironic_capacity_check.py
@@ -54,12 +54,12 @@ def check(auth_ref, args):
 
     if is_up:
         maint_nodes = [node for node in all_nodes if node.maintenance]
+        maint_nodes_count = len(maint_nodes)
         total_nodes = len(all_nodes)
-        up_nodes = total_nodes - len(maint_nodes)
-        # question: is capacity the maint_nodes_percentage
-        # or 1 - maint_nodes_percentage ?
-        metric('ironic_total_nodes_count', 'uint32', total_nodes)
+        up_nodes = total_nodes - maint_nodes_count
+        metric('ironic_maint_nodes_count', 'uint32', maint_nodes_count)
         metric('ironic_up_nodes_count', 'uint32', up_nodes)
+        metric('ironic_total_nodes_count', 'uint32', total_nodes)
 
 
 def main(args):
--- a/playbooks/files/rax-maas/plugins/ironic_capacity_check.py
+++ b/playbooks/files/rax-maas/plugins/ironic_capacity_check.py
@@ -57,7 +57,6 @@ def check(auth_ref, args):
         maint_nodes_count = len(maint_nodes)
         total_nodes = len(all_nodes)
         up_nodes = total_nodes - maint_nodes_count
-        metric('ironic_maint_nodes_count', 'uint32', maint_nodes_count)
         metric('ironic_up_nodes_count', 'uint32', up_nodes)
         metric('ironic_total_nodes_count', 'uint32', total_nodes)
 
--- a/playbooks/files/rax-maas/plugins/ironic_capacity_check.py
+++ b/playbooks/files/rax-maas/plugins/ironic_capacity_check.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python
+
+# Copyright 2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import time
+
+import ipaddr
+from ironicclient import exc
+from maas_common import get_auth_ref
+from maas_common import get_ironic_client
+from maas_common import metric
+from maas_common import metric_bool
+from maas_common import print_output
+from maas_common import status_err
+from maas_common import status_ok
+
+
+def check(auth_ref, args):
+
+    IRONIC_ENDPOINT = ('http://{ip}:6385/v1'.format(ip=args.ip))
+
+    try:
+        if args.ip:
+            ironic = get_ironic_client(endpoint=IRONIC_ENDPOINT)
+        else:
+            ironic = get_ironic_client()
+
+        is_up = True
+
+    except exc.ClientException:
+        is_up = False
+    # Any other exception presumably isn't an API error
+    except Exception as e:
+        metric_bool('client_success', False, m_name='maas_ironic')
+        status_err(str(e), m_name='maas_ironic')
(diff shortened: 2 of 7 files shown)


playbooks/templates/rax-maas/ironic_api_local_check.yaml.j2
```diff
@@ -9,12 +9,12 @@ details     :
     file    : run_plugin_in_venv.sh
     args    : ["{{ maas_plugin_dir }}/ironic_api_local_check.py", "{{ ansible_host }}"]
 alarms      :
-    ironic_api_local_status :
-        label                   : ironic_api_local_status--{{ inventory_hostname }}
+    ironic_capacity :
+        label                   : ironic_capacity--{{ inventory_hostname }}
         notification_plan_id    : "{{ maas_notification_plan }}"
```
npawelek: This has changed in rcbops/master.

playbooks/templates/rax-maas/ironic_capacity_check.yaml.j2
```diff
@@ -0,0 +1,20 @@
...
+period      : "{{ maas_check_period_override[label] | default(maas_check_period) }}"
+timeout     : "{{ maas_check_timeout_override[label] | default(maas_check_timeout) }}"
+disabled    : "{{ (check_name | match(maas_excluded_checks_regex)) | ternary('true', 'false') }}"
+details     :
+    file    : run_plugin_in_venv.sh
+    args    : ["{{ maas_plugin_dir }}/ironic_capacity_check.py", "{{ ansible_host }}"]
+alarms      :
+    ironic_api_local_status :
+        label                   : ironic_api_local_status--{{ inventory_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
```
npawelek: This has changed in rcbops/master.

playbooks/templates/rax-maas/ironic_api_local_check.yaml.j2
tonytan4ever: This is a file name mismatch. Addressed.

playbooks/templates/rax-maas/ironic_capacity_check.yaml.j2
tonytan4ever: This is a file name mismatch. Addressed.

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
```diff
@@ -0,0 +1,83 @@
...
+            ironic = get_ironic_client()
+
+        is_up = True
+
+    except exc.ClientException:
+        is_up = False
+    # Any other exception presumably isn't an API error
+    except Exception as e:
+        metric_bool('client_success', False, m_name='maas_ironic')
+        status_err(str(e), m_name='maas_ironic')
```
supermari0: Don't you need to return here? Otherwise, you'll call status_err and status_ok on L54.

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
```diff
@@ -0,0 +1,83 @@
...
+    try:
+        if args.ip:
+            ironic = get_ironic_client(endpoint=IRONIC_ENDPOINT)
+        else:
+            ironic = get_ironic_client()
+
+        is_up = True
+
+    except exc.ClientException:
+        is_up = False
```
supermari0: Does a non-200 response throw a ClientException?

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
```diff
@@ -0,0 +1,83 @@
...
+    # Any other exception presumably isn't an API error
+    except Exception as e:
+        metric_bool('client_success', False, m_name='maas_ironic')
+        status_err(str(e), m_name='maas_ironic')
+    else:
+        metric_bool('client_success', True, m_name='maas_ironic')
+        # pass limit=0 to list all nodes list without pagination
+        all_nodes = ironic.node.list(limit=0)
+
+    status_ok(m_name='maas_ironic')
```
supermari0: Why do you call status_ok if is_up is False?

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
```diff
@@ -0,0 +1,83 @@
...
+    if is_up:
+        maint_nodes = [node for node in all_nodes if node.maintenance]
+        maint_nodes_percentage = len(maint_nodes)/len(all_nodes)
+        # question: is capacity the maint_nodes_percentage
+        # or 1 - maint_nodes_percentage ?
```
supermari0: 1 - maint_nodes_percentage. It doesn't matter as long as you alert on the correct % though

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
```diff
@@ -0,0 +1,83 @@
...
+        status_err(str(e), m_name='maas_ironic')
+    else:
+        metric_bool('client_success', True, m_name='maas_ironic')
+        # pass limit=0 to list all nodes list without pagination
+        all_nodes = ironic.node.list(limit=0)
+
+    status_ok(m_name='maas_ironic')
+    if is_up:
+        maint_nodes = [node for node in all_nodes if node.maintenance]
+        maint_nodes_percentage = len(maint_nodes)/len(all_nodes)
```
supermari0: Is this going to be used in a python3 runtime? If not, you need to cast one or both of these to a float.

In python 2.7:
>>> len([1,2,3]) / len([1,2,3,4])
0

3.5:

>>> len([1,2,3])/len([1,2,3,4])
0.75

playbooks/templates/rax-maas/ironic_capacity_check.yaml.j2
```diff
@@ -0,0 +1,20 @@
+{% set label = "ironic_api_local_check" %}
```
supermari0: Let's get someone more familiar with MaaS checks to thoroughly review this.

Also, do we need to support private network checks? cc @briancurtin

playbooks/templates/rax-maas/ironic_capacity_check.yaml.j2
```diff
@@ -0,0 +1,20 @@
...
+    file    : run_plugin_in_venv.sh
+    args    : ["{{ maas_plugin_dir }}/ironic_api_local_check.py", "{{ ansible_host }}"]
+alarms      :
+    ironic_capacity :
+        label                   : ironic_capacity--{{ inventory_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan_override[label] | default(maas_notification_plan) }}"
+        disabled                : {{ (('ironic_capacity--'+inventory_hostname) | match(maas_excluded_alarms_regex)) | ternary('true', 'false') }}
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (metric["ironic_capacity"] <= 20) {
```
supermari0: Maybe we should make 20 a variable. Not sure what the pattern about that is for this repo though.

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
briancurtin: Yeah, need to make sure we don't report error but also drop through into ok. You could probably just indent L54 and this would be fine.

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
briancurtin: It's currently running 2.7.6, at least on this 14.04 I just looked at. Would make sense to make sure this works in both spots, though.

briancurtin: One thing to note is that the check itself can do the percentage, so you could output a metric for `len(maint_nodes)` and a metric for `len(all_nodes)` and then in the check itself do `percentage(maint_nodes, all_nodes)`—I just did that on a check for file descriptor usage in the maas poller itself, and I got the idea to do that from the conntrack check, so it seems like others are also going that way. 

I don't think what you're doing is "wrong" but it may be easier to just have this plugin export the data and have the check figure out what to calculate from it.

playbooks/templates/rax-maas/ironic_capacity_check.yaml.j2
briancurtin: If you look at [playbooks/vars/maas-openstack.yml](https://github.com/rcbops/rpc-maas/blob/master/playbooks/vars/maas-openstack.yml) there are some variables configured for thresholds. Would probably make sense to get this in there.

briancurtin: Here's what I was referring to earlier with letting the check calculate the percentage itself. [This check](https://github.com/rcbops/rpc-maas/blob/master/playbooks/templates/rax-maas/conntrack_count.yaml.j2#L18) is an example of what I was talking about.

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
tonytan4ever: Yes,  by looking at:
https://github.com/openstack/python-ironicclient/blob/69b4dda421ec18b8dc61bb0396fd38da7498f746/ironicclient/common/apiclient/exceptions.py#L33

All 4xx, 5xx responses will throw subclasses of Client exception. I think that's a pretty comprehensive list.

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
tonytan4ever: This is done.

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
tonytan4ever: This is done. Should only call status_ok if there is no exception.

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
tonytan4ever: Good to know.  This is addressed.

tonytan4ever: This is done. Percentage now is actually calculated in the check:
ironic_capacity_check.yaml.j2

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
tonytan4ever: This is done. Percentage now is actually calculated in the check:
ironic_capacity_check.yaml.j2

playbooks/templates/rax-maas/ironic_capacity_check.yaml.j2
tonytan4ever: This is the wrong label. It is addressed.

playbooks/templates/rax-maas/ironic_capacity_check.yaml.j2
tonytan4ever: Adopted @briancurtin 's comment. Added threshold config variables and let the check calculate percentages.

playbooks/files/rax-maas/plugins/README.md
```diff
@@ -161,6 +161,27 @@ IP address of service to test
...
+##### Description:
+Polls a (native) ironic API living on the specified IP. Gather the number of total nodes and up nodes (nodes not in maintenance)
+
+##### Mandatory Arguments:
+IP address of service to test
+##### Example Output:
+
+    status okay
+    metric client_success uint32 1
+<<<<<<< Upstream, based on branch 'TURTLES-106_ironic_capacity_check' of https://github.com/rcbops/rpc-maas.git
```
briancurtin: FYI you have some merge stuff in here

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
```diff
@@ -0,0 +1,82 @@
...
+        status_ok(m_name='maas_ironic')
+
+    if is_up:
+        maint_nodes = [node for node in all_nodes if node.maintenance]
+        maint_nodes_count = len(maint_nodes)
+        total_nodes = len(all_nodes)
+        up_nodes = total_nodes - maint_nodes_count
+        metric('ironic_maint_nodes_count', 'uint32', maint_nodes_count)
+        metric('ironic_up_nodes_count', 'uint32', up_nodes)
+        metric('ironic_total_nodes_count', 'uint32', total_nodes)
```
briancurtin: Cool, I think this style looks a little better.

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
```diff
@@ -0,0 +1,82 @@
...
+        # pass limit=0 to list all nodes list without pagination
+        all_nodes = ironic.node.list(limit=0)
+        status_ok(m_name='maas_ironic')
+
+    if is_up:
+        maint_nodes = [node for node in all_nodes if node.maintenance]
+        maint_nodes_count = len(maint_nodes)
+        total_nodes = len(all_nodes)
+        up_nodes = total_nodes - maint_nodes_count
+        metric('ironic_maint_nodes_count', 'uint32', maint_nodes_count)
```
briancurtin: It looks like the check is only using `up_nodes` and `total_nodes`—do we need to be sending `maint_nodes_count` here as well?

tonytan4ever: Yes, this is done

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
supermari0: 👍

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
supermari0: 👍

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
supermari0: 👍

playbooks/files/rax-maas/plugins/ironic_capacity_check.py
cjloader: looks good.

playbooks/templates/rax-maas/ironic_capacity_check.yaml.j2
```diff
@@ -0,0 +1,23 @@
...
+        notification_plan_id    : "{{ maas_notification_plan_override[label] | default(maas_notification_plan) }}"
+        disabled                : {{ (('ironic_capacity--'+inventory_hostname) | match(maas_excluded_alarms_regex)) | ternary('true', 'false') }}
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (percentage(metric["ironic_up_nodes_count"] , metric["ironic_total_nodes_count"]) <= {{ maas_capacity_percent_critical_threshold }}) {
+                return new AlarmStatus(CRITICAL, "Bare-metal capacity has dropped to <= {{ maas_capacity_percent_critical_threshold }}%");
+            }
+            if (percentage(metric["ironic_up_nodes_count"] , metric["ironic_total_nodes_count"]) <= {{ maas_capacity_percent_warning_threshold }}) {
+                return new AlarmStatus(WARNING, "Bare-metal capacity has dropped to <= {{ maas_capacity_percent_warning_threshold }}%");
+            }
```
cjloader: looks good

playbooks/templates/rax-maas/ironic_capacity_check.yaml.j2
```diff
@@ -0,0 +1,23 @@
...
+    file    : run_plugin_in_venv.sh
+    args    : ["{{ maas_plugin_dir }}/ironic_capacity_check.py", "{{ ansible_host }}"]
+alarms      :
+    ironic_capacity :
+        label                   : ironic_capacity--{{ inventory_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan_override[label] | default(maas_notification_plan) }}"
+        disabled                : {{ (('ironic_capacity--'+inventory_hostname) | match(maas_excluded_alarms_regex)) | ternary('true', 'false') }}
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (percentage(metric["ironic_up_nodes_count"] , metric["ironic_total_nodes_count"]) <= {{ maas_capacity_percent_critical_threshold }}) {
```
briancurtin: What I meant with my previous comment was that `ironic_maint_nodes_count` gets output from the plugin as a metric but nothing in here consumes it—we only look at up versus total. It seems like maybe we shouldn't be reporting that as a metric.

tonytan4ever: This is addressed

PR #331, opened 2017-08-03
Add ironic-conductor check

tonytan4ever:


playbooks/files/rax-maas/plugins/ironic_conductor_check.py
```diff
@@ -0,0 +1,50 @@
...
+
+from maas_common import get_auth_ref
+from maas_common import metric_bool
+from maas_common import print_output
+
+
+def check(auth_ref, args):
+    name = "ironic-conductor_status"
+    for proc in psutil.process_iter():
+        if 'ironic-conducto' in proc.name():
```
briancurtin: This is working because the beginning of it matches the name, but it'd be nicer to have the full spelling in there. Note the missing R on the end.

tonytan4ever: @briancurtin:  It actually is not a misspelling.  I run this script on  beta environment  and the ironic-coductor process's name is indeed **ironic-conducto**.   Weird as it is I agree.

See this snippet:
https://gist.github.com/tonytan4ever/51ee331b999518f6d1aaa4b3a0c358e6

briancurtin: Hah, weird! My apologies, this is fine then 👍

PR #335, opened 2017-08-21
RE-312 Implement overridable vars for influx repo

odyssey4me:
In order to ensure that RPC-O deployments make use
of a frozen set of apt artifacts, the repo/key
settings must be overridable.

While it'd be ideal for the variables to be set as
role defaults, allowing these settings to be set
via group_vars - this is not possible due to the
way playbooks have been used instead of roles.

As such, we use vars and RPC-O will be forced to
use extra-vars as that's the only option which
has a higher precedence.


odyssey4me:
recheck_cit_trusty_newton

PR #336, opened 2017-08-22
Change affinity to 1 for MaaS Testing

andymcc:
We don't need to build a galera or rabbit cluster, and we aren't testing
multiple Keystone/Horizon/Repo servers.

This patch changes the affinity to "1" for each of the services to avoid
multiple containers being built for the same service, and to avoid a big
in the EOL'd Liberty branch causing issues with mysql connections.

We can't use config overrides as there is a bug in liberty that means
integers cant be interpreted properly - so we have to sed the repo
directly.


andymcc:
recheck_cit_trusty_liberty

andymcc:
Failed due to influxdb repo.

PR #338, opened 2017-08-24
Rearrange poller restart for standalone ceph

briancurtin:
Standalone ceph deploys are currently erroring out as rackspace-monitoring-poller won't be deployed there, but the restart task for the poller first checks for `groups['shared-infra_hosts']`.
That group doesn't exist in the stripped down inventory where this is being run, so flipping the ordering to first check for the poller's deployment will keep us from continuing on in the `when` checks for things like that group which might not be available.

### DIFF
--- a/playbooks/maas-restart.yml
+++ b/playbooks/maas-restart.yml
@@ -27,8 +27,8 @@
       retries: 3
       delay: 2
       when:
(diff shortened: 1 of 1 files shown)


briancurtin:
@cloudnull you able to look at this when you get a chance? This would probably be in 1.1.1?

PR #344, opened 2017-09-05
Cephplusrgw

letterj:
Addition of rgw montioring for ceph Rados Gateway.

This adds general monitoring for nodes in the rgws group.
Adds a local check that hits the rgw endpoint    ( http://localhost:<port>/)  and returns:
0 = Failure to Connect or Other failure    (Critical)
1 = Status Code > 299 (WARNING)
2 = Status Code >= 200 < 299 .    (OK)

Tested it on the rpc-CephAIO
Entities Added for test
cephrgw1
cephrgw2


cloudnull:
recheck_cit_xenial_newton

andymcc:
@cloudnull we're looking to add an rpc-ceph repo that will do some testing - we have a working setup, we need a bit of change to fix that up and get it officially involved.

That said, I'm not too sure the best way of kicking off a ceph build only on ceph changes - im sure we can work that side of it out.

letterj:
@cloudnull   Can you clarify your comment "plan to get this tested"?     I deployed this on a CephAIO in the Rackspace Public Cloud, set up entities for all the containers and the all the checks ran successfully.      I looked around for any other type of MaaS testing and didn't  see it.      Do you mean testing ceph?

cloudnull:
recheck_cit_xenial_pike

cloudnull:
@letterj The tests for this repo are in the tests directory [ https://github.com/rcbops/rpc-maas/tree/master/tests ]. These tests are executed via the CIT jenkins which you can see in the code review section of this PR. At this time we've no automated tests for ceph. If you're testing in an AIO that's great but its meaningless in terms of ensuring we're not introducing regressions or breaking this code path by working on other aspects of the code base. I would recommend getting this repo to use the cephAIO work you've mentioned and adding an **IRR_CONTEXT** to the gating repo so we can then begin testing our ceph monitoring capabilities on a per-commit basis.

cloudnull:
@letterj I suspect you can simply add **ceph** into the list here [ https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/irr_role_test.yml#L19-L27 ] and then add a conditional here [ https://github.com/rcbops/rpc-maas/blob/master/tests/aio-create.sh#L97 ] to run your cephAIO tools when the **IRR_CONTEXT** is set to **ceph**. Once the cephAIO is being used here and we can exercise maas ceph code path I think we'll be good.

andymcc:
recheck_cit_trusty_mitaka

playbooks/templates/rax-maas/ceph_rgw_stats.yaml.j2
```diff
@@ -0,0 +1,30 @@
...
+type        : agent.plugin
+label       : "{{ check_name }}"
+period      : "{{ maas_check_period_override[label] | default(maas_check_period) }}"
+timeout     : "{{ maas_check_timeout_override[label] | default(maas_check_timeout) }}"
+disabled    : "{{ (check_name | match(maas_excluded_checks_regex)) | ternary('true', 'false') }}"
+details     :
+    file    : run_plugin_in_venv.sh
+    args    : {{ ceph_args }}
+alarms      :
+    ceph_rgw_check.{{ ansible_host }} :
```
npawelek: Can we change `ceph_rgw_check` to `ceph_rgw_status` and remove entries for `.{{ ansible_host }}` on L19, L20, and L22?

playbooks/maas-ceph-rgw.yml
```diff
@@ -0,0 +1,70 @@
...
+
+- include: maas-ceph-raxmon.yml
+
+- name: Install checks for ceph rgws
+  hosts: rgws
+  gather_facts: false
+  pre_tasks:
+    - name: Write Ceph monitoring client key to file
+      copy:
+        content: "{{ [hostvars[groups['mons'][0]]['ceph_raxmon_client']['stdout'], '\n'] | join('') }}"
```
letterj: Instead of this being "[hostvars[groups['mons'][0]]"  shouldn't it be "[hostvars[groups['rgws']]

playbooks/maas-ceph-rgw.yml
```diff
@@ -0,0 +1,70 @@
...
+        dest: "/etc/ceph/ceph.client.raxmon.keyring"
+      delegate_to: "{{ physical_host | default(ansible_host) }}"
+
+    - name: Write Ceph monitoring client key to file within legacy rpc hosts
+      copy:
+        content: "{{ [hostvars[groups['mons'][0]]['ceph_raxmon_client']['stdout'], '\n'] | join('') }}"
```
letterj: Instead of this being "[hostvars[groups['mons'][0]]"  shouldn't it be "[hostvars[groups['rgws']]

letterj:
Taking things in a different direction.

PR #345, opened 2017-09-05
Fix galera template variables

npawelek:
The galera check template did not have properly terminated variables. They have now been updated to display their associated values.


cloudnull:
recheck_cit_xenial_ocata

PR #346, opened 2017-09-05
Change alarm criteria variable ansible_host to inventory_hostname

npawelek:
The template is currently generated with an IP address, which can be a bit confusing for support when a ticket is generated. Updated the variable to `inventory_hostname` instead so that identification is easier.


cloudnull:
recheck_cit_trusty_liberty

PR #348, opened 2017-09-07
HUM-304 Allow the use of other ansible-playbook binaries

andymcc:
For non OpenStack deploys we shouldnt enforce the use of
openstack-ansible. For example, for deploying Ceph standalone.

We should allow MaaS to not force the use of openstack-ansible binary
and instead allow this to be set by an environment variable.


andymcc:
recheck_cit_trusty_mitaka

andymcc:
recheck_cit_trusty_mitaka

andymcc:
external repo failure

andymcc:
recheck_cit_trusty_mitaka

PR #349, opened 2017-09-07
HUM-304 Monitor Ceph OSD service in containers

andymcc:
For testing multiple OSD hosts in an AIO we need to run these in
containers, we need to target OSD MaaS against the container correctly,
this patch allows the container_name to be passed to ceph_monitoring.py.

### DIFF
--- a/playbooks/templates/rax-maas/ceph_osd_stats.yaml.j2
+++ b/playbooks/templates/rax-maas/ceph_osd_stats.yaml.j2
@@ -1,6 +1,9 @@
 {% set label = "ceph_osd_stats" %}
 {% set check_name = label+'--'+inventory_hostname %}
 {% set ceph_args = [maas_plugin_dir + "/ceph_monitoring.py", "--name", "client.raxmon", "--keyring", "/etc/ceph/ceph.client.raxmon.keyring"] %}
+{% if physical_host is defined and physical_host != ansible_host %}
+{%   set _ = ceph_args.extend(["--container-name", container_name]) %}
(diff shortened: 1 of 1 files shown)


playbooks/templates/rax-maas/ceph_osd_stats.yaml.j2
```diff
@@ -1,6 +1,9 @@
 {% set label = "ceph_osd_stats" %}
 {% set check_name = label+'--'+inventory_hostname %}
 {% set ceph_args = [maas_plugin_dir + "/ceph_monitoring.py", "--name", "client.raxmon", "--keyring", "/etc/ceph/ceph.client.raxmon.keyring"] %}
+{% if physical_host is defined and physical_host != ansible_host %}
+{%   set _ = ceph_args.extend(["--container-name", container_name]) %}
```
briancurtin: Approved since this works and is fine, but a question: do you even need the `set _ =` part of it? Seems like it could also be `{{ ceph_args.extend(["--container-name", container_name]) }}`?

andymcc: Yeah you'd think - I believe it doesn't work that way (unless the newer versions of jinja2 do this), so you have to do a dummy set for it to actually run the ceph_args.extend. The set bit doesnt really do anything though.

andymcc:
recheck_cit_xenial_newton

andymcc:
recheck_cit_xenial_ocata

PR #350, opened 2017-09-07
HUM-304 Install sysstat for disk_utilisation.py

andymcc:
disk_utilisation.py requires sysstat for iostat to be present. In RPC-O
the openstack-hosts role will install sysstat so for RPC-O it was always
present.

When using a standlone ceph deployment or other service that doesn't
require the openstack-hosts role, we need sysstat in order to run the
disk_utilisation.py monitoring script.


andymcc:
recheck_cit_trusty_liberty

PR #351, opened 2017-09-08
Fix using a list for packages instead of with_items

andymcc:
When using the apt module you can pass a list, but not separated by
spaces, it should just be the list var. The current package install for
maas-poller will break, this patch fixes that and makes the maas-agent
packages install in the same, more efficient, way - that prevents the
need for a task per package.

### DIFF
--- a/playbooks/common-tasks/maas-poller-ubuntu-install.yml
+++ b/playbooks/common-tasks/maas-poller-ubuntu-install.yml
@@ -36,7 +36,8 @@
 
 - name: Ensure container packages are installed
   apt:
-    pkg: "{{ maas_poller_distro_packages | join(' ') }}"
+    pkg: "{{ item }}"
     state: "{{ maas_package_state }}"
     update_cache: yes
     cache_valid_time: 600
+  with_items: "{{ maas_poller_distro_packages }}"


andymcc:
Found this bug trying to implement the same for: https://github.com/rcbops/rpc-maas/pull/350
It fails because it can't install a list like that - we either need to use the package module (which I'd say is the better approach, it will accept a list) or use with_items with the apt module.

playbooks/common-tasks/maas-poller-ubuntu-install.yml
```diff
@@ -36,7 +36,7 @@
 
 - name: Ensure container packages are installed
   apt:
-    pkg: "{{ maas_poller_distro_packages | join(' ') }}"
+    pkg: "{{ maas_poller_dstro_packages }}"
```
odyssey4me: Spelling error.

playbooks/common-tasks/maas-agent-ubuntu-install.yml
```diff
@@ -46,8 +46,7 @@
 
 - name: Ensure container packages are installed
   apt:
-    pkg: "{{ item }}"
+    pkg: "{{ maas_distro_packages }}"
```
odyssey4me: Passing a list won't work with Ansible < 2.2 and this set of playbooks tries to cater for 1.9+ so we will have to use a with_items loop.

andymcc:
recheck_continuous-integration_travis-ci_push

PR #352, opened 2017-09-12
Fix issue with filesystem template population

npawelek:
Template was improperly deploying as a result of not specifying the value of `item.filesystem`


npawelek:
recheck_cit_trusty_liberty

PR #353, opened 2017-09-12
Deploy holland by default

npawelek:
Holland was depending on a virtualenv that didn't exist. I've changed this to deploy based on a boolean instead. Additionally, fixed a few typos in the holland plugin and template.


BjoernT:
That problem should not exist if you deploy RPC-O which does come with openstack-ops and that setups the venv for holland via https://github.com/rsoprivatecloud/openstack-ops/blob/master/tasks/holland_config.yml, or in older releases the support role for RPC-O
This indicated the deployment was tested on a stock OSA release ?

npawelek:
Closing for the time being

PR #354, opened 2017-09-13
Fix issue preventing Holland from deploying on all supported rpc-o versions

npawelek:
Holland was only configured to deploy against newton with the new virtualenv configuration. This change enables the ability to determine if Holland should be deployed at all, and if a virtualenv should be used.


npawelek:
recheck_cit_trusty_mitaka

npawelek:
recheck_cit_trusty_kilo

npawelek:
recheck_cit_trusty_mitaka

PR #355, opened 2017-09-20
[DNM] Test Ceph /w MaaS build

andymcc:


andymcc:
recheck_cit_all

andymcc:
This works now - we'll need to rework how we implement this - intended this patch to show how it could work though.

PR #358, opened 2017-09-28
RO-2619: Change alarm label to be consistent with other alarms

npawelek:
Fix alarm label to align naming across all templates.


npawelek:
recheck_cit_all

PR #359, opened 2017-10-02
Update notification plans allowing everything to be overridden

npawelek:
Some alarm definitions didn't include the override option for changing the notification plan. This ensures all templates have the ability to do so.


npawelek:
recheck_cit_all

PR #360, opened 2017-10-02
RE-613 Respect artifact location environment vars

hughsaunders:
RE provide environment variables which hold the path to the
correct place to store artifacts and test results:
  * RE_HOOK_ARTIFACT_DIR
  * RE_HOOK_RESULT_DIR

These are now available in the IRR jobs so can be used here.

Depends: https://github.com/rcbops/rpc-gating/pull/450


hughsaunders:
recheck_cit_all various failures - excluded checks present in api results, errors creating temp files. None look related.

PR #369, opened 2017-10-13
Octavia process and api checks

xgerman:


cloudnull:
change looks good however its currently failing lint.

npawelek:
@xgerman @cloudnull This is missing a maas scheme specific to octavia as well as a standard lb_api_check (not private_lb_api).

PR #370, opened 2017-10-18
RO-2832 Fix very-maas spacing

hughsaunders:
Previously there were no spaces between --excluded check arguments
which lead to a failure to parse the arguments correctly. This
commit corrects that.


hughsaunders:
recheck_cit_trusty_liberty recheck_cit_trusty_newton

PR #375, opened 2017-10-21
RO-2908 Set correct check port for novnc enabled console

BjoernT:
Use the correct HTTP port 6080 for the nova console novnc


BjoernT:
Issues as a result of hard set rpc-maas values from the Kilo rollout.
Those are no longer applied during leapfrog upgrade, hence the issue is resolved

PR #377, opened 2017-10-26
Fix some hard-coded http endpoints

supermari0:
Fixes hard-coded HTTP endpoints for the following checks:
- nova service checks
- nova cloud stats checks
- cinder service checks
- neutron service checks
- ceph radosgw checks
- rabbitmq checks

This also makes the nova cloud stats hit the VIP instead of the ansible
host so proper TLS termination can occur in that case.

### DIFF
--- a/playbooks/files/rax-maas/plugins/ceph_monitoring.py
+++ b/playbooks/files/rax-maas/plugins/ceph_monitoring.py
@@ -39,10 +39,11 @@ def check_command(command, container_name=None):
     return json.loads(lines[-1])
 
 
-def get_ceph_rgw_hostcheck(rgw_host, rgw_port, container_name=None):
-    host_url = "http://%s:%s" % (rgw_host, rgw_port)
+def get_ceph_rgw_hostcheck(rgw_protocol, rgw_host, rgw_port,
+                           container_name=None):
+    host_url = "%s://%s:%s" % (rgw_protocol, rgw_host, rgw_port)
     try:
-        sc = requests.get(host_url).status_code
+        sc = requests.get(host_url, verify=False).status_code
         if (sc >= 200) and (sc < 300):
             status_code = 2
         else:
@@ -87,9 +88,9 @@ def get_mon_statistics(client=None, keyring=None, host=None,
     maas_common.metric('mon_health', 'uint32', health_status)
 
 
-def get_rgw_checkup(client, keyring=None, rgw_port=None,
+def get_rgw_checkup(client, keyring=None, rgw_protocol=None, rgw_port=None,
                     rgw_host=None, container_name=None):
-    rgw_status = get_ceph_rgw_hostcheck(rgw_host, rgw_port,
+    rgw_status = get_ceph_rgw_hostcheck(rgw_protocol, rgw_host, rgw_port,
                                         container_name=container_name)
     maas_common.metric('rgw_up', 'uint32', rgw_status)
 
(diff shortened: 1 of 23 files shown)


npawelek:
@cloudnull If you get a second, do you mind reviewing this?

cloudnull:
recheck_cit_all

playbooks/vars/main.yml
```diff
@@ -265,3 +265,37 @@ maas_holland_enabled: false
 #                            in the galera containers (RPC-O newton and onward).
 #
 maas_holland_venv_enabled: false
+
+#
+# nova_service_check_protocol: Protocol to use when authenticating for nova
+#                              service checks. Valid values are "http" or
+#                              "https".
+# TODO Should this be moved to the openstack vars file?
```
cloudnull: it makes sense that this is in the main vars as it will be shared across various services and it's nice to only have to modify or look for these vars in a single place.

playbooks/files/rax-maas/plugins/ceph_monitoring.py
```diff
@@ -39,10 +39,11 @@ def check_command(command, container_name=None):
...
 
 
-def get_ceph_rgw_hostcheck(rgw_host, rgw_port, container_name=None):
-    host_url = "http://%s:%s" % (rgw_host, rgw_port)
+def get_ceph_rgw_hostcheck(rgw_protocol, rgw_host, rgw_port,
+                           container_name=None):
+    host_url = "%s://%s:%s" % (rgw_protocol, rgw_host, rgw_port)
     try:
-        sc = requests.get(host_url).status_code
+        sc = requests.get(host_url, verify=False).status_code
```
cloudnull: I'm personally not a fan of force cert acceptance, but I think we can follow on later to resolve this if we need.

playbooks/vars/main.yml
supermari0: ok, will move - I wasn't sure whether those vars files actually got used, so I stuck it in a place where I knew the templates would find it.

cloudnull:
looks good to me, we just need to make CIT happy now.

playbooks/vars/main.yml
supermari0: oh, never mind - I'll remove the TODO

playbooks/files/rax-maas/plugins/ceph_monitoring.py
supermari0: ditto - I heard that we use forced cert acceptance in some other calls, so I presumed there was a reason.

I'm making a follow up card in jira today to:

1. fix the rest of the hard-coded values (including hard-coded ports)
2. add a var where you can put cert/ca info if it's not a CA that can be verified by the system itself

CrashenX: Me either... @supermari0 / @briancurtin can you make a future work follow on task for @dusty-simoni to review.

supermari0: 👍

supermari0:
Fixed a typo (s/add_arugment/add_argument/ in ceph_monitoring.py).

Other gate jobs besides ceph were succeeding, so hopefully this will be sufficient.

supermari0:
recheck_cit_all

supermari0:
looks like this was a failure to start the job runner

mattt416:
recheck_cit_trusty_liberty

mattt416:
recheck_cit_trusty_mitaka

mattt416:
recheck_cit_trusty_newton

mattt416:
recheck_cit_xenial_ceph

mattt416:
recheck_cit_xenial_newton

mattt416:
recheck_cit_xenial_ocata

playbooks/files/rax-maas/plugins/ceph_monitoring.py
supermari0: https://rpc-openstack.atlassian.net/browse/TURTLES-313

PR #378, opened 2017-10-26
Add fact to prevent handler failure

npawelek:
This playbook was missing the `maas_current_group` fact which caused the playbook to fail when running the handler.


npawelek:
recheck_cit_all

npawelek:
recheck_cit_all

npawelek:
recheck_cit_trusty_liberty

PR #380, opened 2017-10-27
RE-645 rpc-maas fixes for Phobos Verification

hughsaunders:
These commits fix a few issues discovered while using maas-verify for checking on phobos.

### DIFF
--- a/playbooks/maas-verify.yml
+++ b/playbooks/maas-verify.yml
@@ -22,7 +22,7 @@
       file:
         path: "{{ workspace_logs }}"
         state: "directory"
-        mode: "0755"
+        mode: "1777"
 
     - name: Create tools dir
       file:
--- a/playbooks/maas-verify.yml
+++ b/playbooks/maas-verify.yml
@@ -70,6 +70,7 @@
         minutes: 5
       when:
         - maas_use_api | bool
+        - maas_verify_wait | bool
 
     - name: Run MaaS verify local
       command: "{{ maas_venv_bin }}/python /opt/rpc-maas-tools/rpc-maas-tool.py --logfile {{ workspace_logs }}/rpc-maas-logs.txt verify-local"
--- a/playbooks/maas-verify.yml
+++ b/playbooks/maas-verify.yml
@@ -92,6 +92,7 @@
       delay: 20
       when:
         - maas_use_api | bool
+        - maas_verify_registration | bool
 
     - name: "Verify Check & Alarm Status"
       command: >
--- a/playbooks/maas-verify.yml
+++ b/playbooks/maas-verify.yml
@@ -22,7 +22,6 @@
       file:
         path: "{{ workspace_logs }}"
         state: "directory"
-        mode: "1777"
 
     - name: Create tools dir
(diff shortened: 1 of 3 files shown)


hughsaunders:
recheck_all

hughsaunders:
recheck_cit_all

playbooks/maas-verify.yml
```diff
@@ -22,7 +22,7 @@
       file:
         path: "{{ workspace_logs }}"
         state: "directory"
-        mode: "0755"
+        mode: "1777"
```
mattt416: Maybe it's best to just leave the mode out, so that it doesn't change it if the dir already exists.  The problem w/ using `1777` here is that we don't want some arbitrary dir. to have that perm if the deployer sets `workspace_logs`.

playbooks/vars/main.yml
```diff
@@ -59,7 +59,7 @@ maas_fqdn_extension: ""
 # The entity name is the inventory hostname and fqdn extension combined.
 #
 # Set the hostname for use with the api checks
-maas_entity_name: "{{ inventory_hostname }}{{ maas_fqdn_extension }}"
+maas_entity_name: "{{ ansible_hostname }}{{ maas_fqdn_extension }}"
```
mattt416: Can we get more info on this change here?

hughsaunders: From the commit message:
```
The curently used inventory_hostname does not match the actual
entitiy names. Tied to RE-645 as the problem was discovered while
working on phobos deploy verification job.```

File: playbooks/maas-verify.yml (hunk 1 above)
Comment (hughsaunders) [reply playbooks/maas-verify.yml]: Done

File: playbooks/vars/main.yml (hunk 2 above)
Comment (mattt416) [reply playbooks/vars/main.yml]: OK ... I guess my concern is that this code is being used in prod now, so it's a little unclear why this isn't a problem for other deployments.

**Comment by hughsaunders** [comment]
The trusty mitaka build failed on gathering facts 
```
fatal: [aio1_cinder_api_container-e9558b80] => SSH Error: mux_client_hello_exchange: write packet: Broken pipe
```

recheck_cit_trusty_mitaka

playbooks/vars/main.yml
mattt416: This change make a bit more sense now that I understand that deployers don't always set the inventory_hostname to match their system's hostname (ansible_hostname).

npawelek:
@hughsaunders @mattt416 This was an issue we had to revert a while back as support always deploys based on `{{ inventory_name }}` not `{{ ansible_hostname }}`. This becomes difficult to handle with the unique naming that CDC customers use and how we link naming back to CORE (for standardization with ease of accessing environments). Pulling down future releases of rpc-maas will break many of our existing deployments. We're going to need a way to work around this as it won't work as-is.

hughsaunders:
@npawelek I created [RE-1127](https://rpc-openstack.atlassian.net/browse/RE-1127) to address your comment.

npawelek:
@hughsaunders thank you

PR #383, opened 2017-11-01
[CEPHSTORA-22] Fix MaaS to allow Ceph Luminous support

andymcc:
In order to support luminous, we need to allow access to client.raxmon
for the mgr service (which is required to perform the ceph commands we
run in the script).

Additionally - the health_services stanza from ceph status has been
removed.
https://www.spinics.net/lists/ceph-devel/msg37767.html
This was not a useful measurement and was removed. This patch removes
that check/metric and adjusts the check to use the "mon_in" metric which
will calculate if the mon is in and up.


andymcc:
Just by way of confirmation - I've applied this patch to the Luminous Ceph patch to show that it will work with both Jewel and Luminous: https://github.com/rcbops/rpc-ceph/pull/22

PR #385, opened 2017-11-02
RE-1082 Ensure pyyaml is installed in MaaS venv

hughsaunders:
This is required for rpc-maas-tool, which is used by maas-verify to
check that MaaS functioning correctly.

### DIFF
--- a/playbooks/vars/maas-verify.yml
+++ b/playbooks/vars/maas-verify.yml
@@ -18,3 +18,7 @@
 #
 maas_verify_pip_packages:
   - futures
+  - pyyaml
+  - rackspace-monitoring-cli
+  - requests
+  - waxeye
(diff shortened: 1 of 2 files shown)


playbooks/vars/maas-verify.yml
```diff
@@ -18,3 +18,7 @@
 #
 maas_verify_pip_packages:
   - futures
+  - pyyaml
+  - rackspace-monitoring-cli
+  - requests
+  - waxeye
```
mattt416: Can `waxeye` be removed from playbooks/vars/maas-agent.yml ?  I'd imagine that is only used by rpc-maas-tool.py in the verify.

hughsaunders: Done

PR #388, opened 2017-11-08
RE-1127 Revert "RE-645 use ansible_hostname..."

hughsaunders:
This reverts commit 25f4a298cb67184026d8e1ef289f07cd19ef8fd6.

Reverted because it breaks support's workflow.

### DIFF
--- a/playbooks/vars/main.yml
+++ b/playbooks/vars/main.yml
@@ -59,7 +59,7 @@ maas_fqdn_extension: ""
 # The entity name is the inventory hostname and fqdn extension combined.
 #
 # Set the hostname for use with the api checks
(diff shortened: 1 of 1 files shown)


mattt416:
recheck_cit_trusty_liberty
recheck_cit_trusty_mitaka
recheck_cit_trusty_newton
recheck_cit_xenial_newton
recheck_cit_xenial_ocata

mattt416:
All those jobs failed because we were creating multiple instances which created multiple entities w/ same name, and that caused the maas deployment to fail.

hughsaunders:
recheck_cit_trusty_liberty failed with `Failed to validate the SSL certificate for repos.influxdata.com:443`

PR #391, opened 2017-11-09
INTAKERPC-141:  Add neutron ovs agent check.

tonytan4ever:


npawelek:
@tonytan4ever Looks like there was a discussion in the intake to setup OVS for testing this functionality. I think the only thing that stands out is the methodology for deploying it. Per the initial commit, I don't see an inclusion of the `maas-openstack-neutron.yml` playbook to determine whether `ovs` or `linuxbridge` monitoring is deployed on an environment.

tonytan4ever:
@npawelek:  I have committed new changes to address the concern you have. We can proceed to test now.

npawelek:
@tonytan4ever Thanks. In the context of deploying playbooks on an environment with OVS, I believe the linuxbridge agent check will also deploy as that group still exists. Correct me if I'm wrong, I think we'd only want to do one or the other?

cloudnull:
@npawelek you are correct. We'd only want to deploy an agent verification check for a system in service. So the playbooks should be able to detect the network agent in service and deploy a compatible monitor.

tonytan4ever:
@npawelek:  I added condition for installing linuxbridge agent check, and ovs agent check condition in place. Lmk if there is anything else missing.

npawelek:
recheck_cit_trusty_mitaka

npawelek:
@tonytan4ever looks like this is also failing pep8

npawelek:
recheck_cit_trusty_mitaka

npawelek:
recheck_cit_trusty_liberty

tonytan4ever:
recheck_xenial_master

tonytan4ever:
recheck_cit_xenial_master

PR #393, opened 2017-11-10
Limit agent restarts preventing pollers from going out of sync

npawelek:
This is a temporary workaround to prevent rackspace-monitoring-poller from going out of sync in large environments when the playbooks are ran with a high number of forks. This prevents the need to reach out to the monitoring devs to force a re-sync of the private zone. This has been tested in multiple 200+ node environments and successfully allows all checks to register as intended.

### DIFF
--- a/playbooks/maas-restart.yml
+++ b/playbooks/maas-restart.yml
@@ -16,7 +16,7 @@
 - name: Restart MaaS
   hosts: "{{ maas_restart_on_host_override | default('maas_restart_on_host') }}"
   gather_facts: false
-  serial: 1
+  serial: "{{ maas_restart_serial | default(1) }}"
   tasks:
     - name: Restart rackspace-monitoring-poller (systemd)
       service:
--- a/playbooks/maas-restart.yml
+++ b/playbooks/maas-restart.yml
@@ -16,6 +16,7 @@
 - name: Restart MaaS
   hosts: "{{ maas_restart_on_host_override | default('maas_restart_on_host') }}"
   gather_facts: false
+  serial: 1
   tasks:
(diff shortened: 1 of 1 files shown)


playbooks/maas-restart.yml
```diff
@@ -16,6 +16,7 @@
 - name: Restart MaaS
   hosts: "{{ maas_restart_on_host_override | default('maas_restart_on_host') }}"
   gather_facts: false
+  serial: 1
```
cloudnull: Can we make this configurable or using some percentage for serial instead of forcing all restarts to happen 1 at a time?

maybe using something like:

``` yaml
serial: "{{ maas_restart_serial | default(1) }}"
```

npawelek: Sure, I'll get that pushed out.

npawelek:
@cloudnull Updated.

On a separate note, for the recent deployments of rpc-maas on multiple large clouds (250+ nodes), I think the handler in general is adding 20-30 minutes to the execution time of the playbooks, even before the proposed change to this value. I've been contemplating removing the handler altogether as we should probably restart all agents anytime plays are run. We're already using `maas_restart_independent=false` on every deploy, so I'm wondering if the handler makes sense anymore?

npawelek:
recheck_cit_xenial_newton

cloudnull:
@npawelek Come to think of it, it may make sense to simply have a stand-alone (rolling?) restart playbook that we can just call when needed; if the restart handler is no longer helpful. That said, I'll get this PR merged so we can include this change in the next tag and we can follow on for a future commit to tune this interaction up.

npawelek:
@cloudnull Sounds good, thanks!

PR #395, opened 2017-11-16
Turtles-317:  Add ceph_standalone override options

tonytan4ever:
1. Add ceph_standalone variable, defaults to False.
2. Only deploy private network checks to non ceph_nodes when in Ceph
standalone.


cloudnull:
recheck_cit_trusty_mitaka

cloudnull:
recheck_cit_trusty_liberty

tonytan4ever:
Abandon this PR as per discussion with @npawelek, we will use an empty _shared-infra_hosts_  in ceph-standalone inventory to avoid deployment failure.

PR #396, opened 2017-11-17
Remove all agent restart handlers

npawelek:
In large environments the restart handler adds considerable time, up to 10 minutes, per call to the handler from each playbook (some make multiple calls). This change removes all the dependencies of those handlers allowing the playbooks to iterate quickly and always perform a single restart at the end of `site.yml`. If independent plays are executed, a manual call to `maas-restart.yml` will always perform restarts.


npawelek:
recheck_cit_trusty_mitaka

PR #399, opened 2017-11-28
Add log_file_watcher utility script

tonytan4ever:


briancurtin:
How is this going to be used? There are a few tail-like libraries out there that likely do the same thing, so I'd be interested to know where this is going and what it needs to do.

tonytan4ever:
@briancurtin:  This is a requirement from https://rpc-openstack.atlassian.net/browse/RO-2967
It needs a way to alert ERROR/CRITICAL/WARNING log lines.

There are other tail utility like [pygtail](https://github.com/bgreenlee/pygtail) and [pytailer](pytailer) indeed, here I choose to implement it on own for various reasons:
1. Reduce dependency on extra outside libs, 
2. This one supports a callback mechanism, so we can customize what we want to do upon notifying of the log lines.
3.  This utility also has other features like target file pattern matching, async mode, and specifying # of tail lines listening which are nice

tonytan4ever:
Close for now as we have to find out the broader requirement and available options for catching logs

PR #400, opened 2017-11-28
Use the new HP ssacli over hpssacli

BjoernT:
HP has begun replacing the hpssacli package with ssacli.
The arguments and output format are still the same,
requring minimal effort to make this change.
The HP MCP version is managed as part of the openstack-ops module
and is now frozen to version 10.6 to better synchronize these changes

Closes-Bug: RO-3090

### DIFF
--- a/playbooks/files/rax-maas/plugins/hp_monitoring.py
+++ b/playbooks/files/rax-maas/plugins/hp_monitoring.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import argparse
+import os
 import subprocess
 
 import maas_common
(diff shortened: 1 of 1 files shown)


BjoernT:
I will redo this patch to account for a transition phase where either one or both of the binaries are installed

npawelek:
@BjoernT It does appear that travis-ci is failing lint

BjoernT:
@npawelek I know, it fails in the tigk stuff, so unrelated. I'm looking at these in a different PR as this one is just fine

npawelek:
recheck_cit_trusty_mitaka

PR #401, opened 2017-12-01
Fixing ansible linting errors

BjoernT:
Addressing ansible-lint errors within maas-tigkstack and verify

PR #402, opened 2017-12-04
Add maas_rally performance monitoring

cfarquhar:
This commit adds the following:

- Playbook to deploy performance monitoring
- MaaS plugin to execute and manage rally tasks
- A set of performance scenarios
  - cinder
    - boot instance from local disk
    - create volume
    - attach volume to instance
    - detach volume from instance
    - delete volume
    - delete instance
  - glance
    - create image
    - delete image
  - keystone
    - fetch token
    - validate token
  - neutron_ports
    - create port
    - delete port
  - neutron_secgroups
    - create security group
    - update security group
    - delete security group
  - nova
    - boot instance from local disk
    - delete instance
  - nova_cinder
    - create cinder volume
    - boot instance from volume
    - delete instance
  - swift
    - create container
    - upload object
    - download object
    - delete object
    - delete container

### DIFF
--- a/playbooks/files/rax-maas/plugins/rally_performance.py
+++ b/playbooks/files/rax-maas/plugins/rally_performance.py
@@ -1,0 +1,210 @@
+#!/usr/bin/env python
+
+# Copyright 2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import numpy as np
+import os
+import time
+import yaml
+
+import maas_common
+from maas_common import status_err, status_ok, metric
+
+import rally
+from rally.api import API
+
+PLUGIN_PATH = '/usr/lib/rackspace-monitoring-agent/plugins/rally/plugins/'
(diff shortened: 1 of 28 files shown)


cfarquhar:
Looks like at least some of the gates are fixed now. 

recheck_cit_trusty_mitaka
recheck_cit_trusty_newton
recheck_cit_xenial_ocata

Linting also appears to be fixed in #401.  Anyone know the incantation to re-run that job?

cloudnull:
The change looks great, @cfarquhar can you rebase this PR so that it pulls in the work done to make the gate functional?

cfarquhar:
Thanks for the feedback @cloudnull.

This PR is currently passing all required gates with the exception of `CIT/trusty_liberty` and `CIT/trusty_mitaka`, which will never pass due to dependencies on modern ansible functionality (combine filter, os_* modules, etc).  This PR should not be backported to the `stable/liberty`, or `stable/mitaka` branches.  

Any thoughts on whether we can get this merged as is?

cfarquhar:
Looks like we're passing all required gate tests now.

playbooks/files/rax-maas/plugins/rally_performance.py
```diff
@@ -0,0 +1,210 @@
...
+import numpy as np
+import os
+import time
+import yaml
+
+import maas_common
+from maas_common import status_err, status_ok, metric
+
+import rally
+from rally.api import API
```
tonytan4ever: Question: where did the API module/file come from?  I did not see it in this PR.

cfarquhar: Hi @tonytan4ever - 

The `maas-rally` venv is built [here](https://github.com/rcbops/rpc-maas/pull/402/files#diff-d8f1aacd80c13b8df75f3bad031f07a7R82), which pulls in the rally python module listed [here](https://github.com/rcbops/rpc-maas/pull/402/files#diff-8c21f835e4a002730edc76d76e01711dR205).

We use a separate venv because rally has a bunch of dependencies that we want to isolate from the standard `maas` venv.

PR #404, opened 2017-12-06
Skip ceph rgw stats in docs tool

briancurtin:
The way the docs util handles a recent change to the ceph_rgw_stats check is blocking the ability to build rcbops/privatecloud-docs (as commented in PR #1132). I am overloaded right now and can't fix this, but in order to unblock the docs build, skipping this particular check will allow it to move on.


briancurtin:
Don't bother looking at any gate failures, this doesn't even get executed there.

PR #408, opened 2017-12-08
[WIP] Fix and update the maas-common lib

cloudnull:
The maas_common lib is a gian ball of technical debt which has long been
a dumping ground for code to simply make monitors work. This change
updates the maas-common library to use the openstacksdk and the upstream
monitorstack library. This change is ONLY being implemented in
maas-common and is being developed with care to ensure all of the
plugins work as is without modification. A monitorstack credential
script was created using the original maas-common credential parser, and
the metric generation functions have been broken out into their own file
which helps make it easier to understand what is going within our
monitoring solution.


mattt416:
recheck_cit_xenial_ceph

PR #413, opened 2017-12-12
Skip tox setup for ceph jobs

andymcc:


andymcc:
This could fix the ceph bits but doesnt resolve the trusty issue, closing this in favour of #414

PR #415, opened 2017-12-12
Update maas_influxdata_key variable

mattt416:
In #414 we tried making this include conditional, however this playbook
is still being included when run w/ older versions of ansible (1.9.6
as an example).

Instead, we update maas_influxdata_key to have an id and keyserver key.
This allows apt_key to work since it doesn't need to speak to
https://repos.influxdata.com/ directly.  Using ansible's apt_key with
https://repos.influxdata.com/ fails because of SNI and:

1. https://github.com/ansible/ansible/pull/32053/ (not backported to
   2.3.2.0)
2. ansible 1.9.6 doesn't seem to support SNI on Trusty (python < 2.7.9)

### DIFF
--- a/playbooks/vars/maas-tigkstack.yml
+++ b/playbooks/vars/maas-tigkstack.yml
@@ -17,7 +17,8 @@
 # Set the influxdata distro package source & key
 #
 maas_influxdata_key:
-  url: "https://repos.influxdata.com/influxdb.key"
+  keyserver: "keyserver.ubuntu.com"
+  id: "2582E0C5"
 
 maas_influxdata_repo:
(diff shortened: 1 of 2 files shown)


jedsmith:
This change is probably OK, @CrashenX, because it's just preventing the tests from running. The tests still exist, the way I read it. Happy to be shown I'm wrong.

mattt416:
Hold off on merging this for now -- let's switch the apt_key ansible tasks to using shell commands to bypass the bug ... and we can then let this stuff be deployed/tested as it was.

claco:
I can confirm we'd not like to remove it from an RLM perspective, or at least, we have some work in the future to use it during deployments to measure cp downtime, etc.

tonytan4ever:
@mattt416: Just a quick caveat: switching to shell command might break ansible-lint check

PR #417, opened 2017-12-13
[DNM] Testing with 2 ceph osds

andymcc:
MaaS tests fail with 2 OSDs, but the rpc-ceph repo is working fine for
that when running MaaS.

This PR is to test that and see what failures occur.


andymcc:
recheck_cit_xenial_ceph

PR #428, opened 2018-01-22
[DNM] Troubleshoot gating issues with 'openstack quota set'

cfarquhar:
This commit adds a wait to os_project.  Some gating jobs fail in the next task
("Set quotas for maas_rally projects") with "'NoneType' object has no attribute
'name'" in stderr from the openstack client.  This is testing the hypothesis
that we are attempting to set a quota on the project before it's fully created.


tonytan4ever:
recheck_cit_xenial_newton

PR #429, opened 2018-01-23
Unblock gating

cfarquhar:
Gating is failing due to flakey behavior from the `openstack quota` command
in the `maas-openstack-rally.yml` playbook.  Based on a conversation with
@cloudnull, the behavior is intermittent and the failure occurs even when
the quota has been set appropriately.

Gating will fail in the verification stage if the quotas aren't actually set.


cfarquhar:
Ok @tonytan4ever , all required tests passed.  Can you please merge?

PR #434, opened 2018-01-31
[RO-3877] Reduce protocol variables down to a single variable

npawelek:
The monitoring agent has an arbitrary limit of 10 args per plugin. This change limits ceph_rgw_stats check from 11 args, down to 10.


npawelek:
Turns out the plugin itself uses an argument in the template. In essence, this does not actually fix the issue in entirety. Closing for the time being.

PR #438, opened 2018-01-31
RE-1305 Use matching endpoint with os_ modules

git-harry:
The os_ Ansible tasks are failing because the public (default) endpoint
is https but the certificates are being validated based on whether or
not the admin endpoint is insecure. The admin endpoint is set to http
and so is causing the tests to fail. This change updates the tasks to
ensure the endpoint type is admin and so therefore matching the var used
to determine whether or not the certificates need validating.


git-harry:
recheck_trusty_liberty_deploy

PR #439, opened 2018-02-01
Switches check to use Octavia V2 API

xgerman:
In most installations we will be runnign straight Octavia and hence
Ovtavia V1 AI will be disabled so we need to run checks against
the V2 API.

### DIFF
--- a/playbooks/files/rax-maas/plugins/octavia_api_local_check.py
+++ b/playbooks/files/rax-maas/plugins/octavia_api_local_check.py
@@ -20,6 +20,7 @@ import datetime
 import ipaddr
 from maas_common import get_auth_ref
 from maas_common import get_endpoint_url_for_service
+from maas_common import get_keystone_client
 from maas_common import metric
 from maas_common import metric_bool
 from maas_common import print_output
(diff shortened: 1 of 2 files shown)


xgerman:
recheck

playbooks/files/rax-maas/plugins/octavia_api_local_check.py
```diff
@@ -32,14 +33,20 @@ def check(auth_ref, args):
     OCTAVIA_ENDPOINT = 'http://{ip}:9876/v1'.format(ip=args.ip,)
```
johnsom: This has "/v1" in it.
Ugh, still using the old port # way to access APIs? We should consider moving to the wsgi non-port based deployment.

playbooks/maas-openstack-octavia.yml
```diff
@@ -67,6 +67,12 @@
   hosts: octavia-api
   gather_facts: false
   tasks:
+    - name: Add load-balancer member role
+      shell: ". /root/openrc && openstack role add --project $OS_PROJECT_NAME --user $OS_USERNAME load-balancer_member"
```
johnsom: We should consider the read-only roles for maas: role:load-balancer_observer or role:load-balancer_global_observer

https://docs.openstack.org/octavia/latest/configuration/policy.html

xgerman:

tonytan4ever:
LGTM

PR #440, opened 2018-02-06
[RO-3945] Modify period and timeout overrides to use label instead of check_name

npawelek:
Allows overridability of period and timeout values for these checks.


andymcc:

The bug encountered here is fixed by https://github.com/rcbops/rpc-maas/commit/99c9fc5b402fa1792968e9d08de833852c29d525

PR #443, opened 2018-02-09
CEPHSTORA-116 Exclude nbd devices from disk util checks

andymcc:
nbd devices get generated once the benchmark tool is run as part of
Ceph. This creates devices that are then automatically picked up. We
shouldn't monitor these.

Whilst the devices are unmapped, the /dev/nbdx devices stay around, and
can't be removed.


andymcc:

PR #444, opened 2018-02-09
CEPHSTORA-106 Update ceph_monitoring.py for Luminous

andymcc:
Since Luminous the "overall_status" field in the json output for Ceph
has "HEALTH_WARNING" to alert you to the fact it has changed. With the
following message:

"overall_status":"HEALTH_WARN","detail":["'ceph health' JSON format
has changed in luminous. If you see this your monitoring system is
scraping the wrong fields. Disable this with 'mon health preluminous
compat warning = false'"

We can set the compat warning = false setting, but it would be a better
approach to monitor the correct value.

In order to make this change backwards compatible with Ceph versions
less than Jewel, we will need to use the "overall_status" field when the
"status" field does not exist (e.g. in Jewel).


andymcc:
recheck_cit_container_pep8_tox-test

andymcc:

andymcc:

PR #446, opened 2018-02-13
Parameterize statistic used for Rally alarms

cfarquhar:
This commit makes the following changes parameterizes the description statistic that is used for alarm criteria.  The default metric also changes from '95pctl' to 'mean', which makes more sense for the default setting of concurrency = 1.


cfarquhar:

Unrelated failure:
```
17:22:00 TASK [galera_client : Update apt if needed] ************************************
17:22:04 fatal: [aio1_cinder_api_container-914e1e22]: FAILED! => {"changed": false, "failed": true, "msg": "Could not fetch updated apt files"}
```

PR #451, opened 2018-02-20
Removed old comment

michaelrice:
The comment referenced a bug that had been abandoned ages ago, and was stating the value for a variable was hard coded to False when it was actually hard coded to True since c967c5e2eab6fd8


tonytan4ever:

PR #452, opened 2018-02-20
Move maas_rally plugin config to /etc

cfarquhar:
This change moves the maas_rally plugin configuration from the MaaS agent
plugin directory to /etc.


cfarquhar:
```
aio1_ceilometer_collector_container-787180c6 : ok=8    changed=1    unreachable=1    failed=0   
```

PR #453, opened 2018-02-21
TURTLES-645: Load maas_rally plugin configuration via class

cfarquhar:
This change adds a class to manage plugin configuration in preparation for more
steps in the plugin relying on configuration read from a file.

### DIFF
--- a/playbooks/files/rax-maas/plugins/rally_performance.py
+++ b/playbooks/files/rax-maas/plugins/rally_performance.py
@@ -42,20 +42,39 @@ class CommandNotRecognized(maas_common.MaaSException):
     pass
 
 
-def send_metrics_to_influxdb():
-    with open(PLUGIN_CONF, 'r') as stream:
-        try:
-            plugin_config = yaml.safe_load(stream)
-            influx_config = plugin_config['influxdb']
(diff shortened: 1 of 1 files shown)


playbooks/files/rax-maas/plugins/rally_performance.py
```diff
@@ -248,4 +267,5 @@ def main():
     parser = make_parser()
     args = parser.parse_args()
     with maas_common.print_output(print_telegraf=args.telegraf_output):
+        plugin_config = PluginConfig(PLUGIN_CONF)
```
tonytan4ever: It seems odd as plugin_config is used as a kind of global variable, I think it's better to pass it into the main function main(plugin_config), and in turn passed it in send_metrics_to_influxdb(plugin_config), what do you think?

cfarquhar: Thanks for your feedback.  I've updated the PR as you suggested.

cfarquhar:

cfarquhar:
TURTLES-645

tonytan4ever:
LGTM, merging

PR #455, opened 2018-02-22
TURTLES-623: Monitor bonding stats/status

tonytan4ever:


tonytan4ever:
recheck_all

playbooks/maas-host-network.yml
```diff
@@ -52,3 +77,21 @@
     - vars/maas-host.yml
   tags:
     - maas-hosts-network
+
+- name: Install checks for bonding interfaces
+  hosts: hosts
+  gather_facts: false
+  tasks:
+    - name: Install network throughput checks
```
npawelek: Probably should update this play's name to reflect

playbooks/files/rax-maas/plugins/bonding_iface_check.py
```diff
@@ -0,0 +1,47 @@
...
+
+from maas_common import metric_bool
+from maas_common import print_output
+
+import netiface_check_lib
+
+
+def bonding_ifaces_check():
+    bonding_ifaces = os.lisdir("/proc/net/bonding")
+    for bonding_iface in bonding_ifaces:
```
npawelek: I've noticed that some environments might have a bond file located there, however, it's empty and contains no sub-interfaces. Will this fail to iterate over causing an alarm or just negate the output altogether?

tonytan4ever: This would just produce no metric (skip the iteration) and generate no alarns

tonytan4ever:

tonytan4ever:

npawelek:
@tonytan4ever Can we get this merged today?

PR #471, opened 2018-03-19
TURTLES-313 add ssl support to magnum api checks

michaelrice:
* Adding protocol and port option to local magnum api check.
* Renamed local variable in function to be all lower case per pep8 standards.

add ssl support to magnum_service_check

* Adding protocol and port option to local magnum api check.
* Renamed local variable in function to be all lower case per pep8 standards.


tonytan4ever:
recheck_trusty_mitaka_deploy

tonytan4ever:
We are not using magnum monitoring anymore,  closing for now.

PR #473, opened 2018-03-19
TURTLES-313 add ssl support to nova_api_local_check

michaelrice:
* Adding protocol and port option to local nova api check.
* Renamed local variable in function to be all lower case per pep8 standards.

add ssl support to nova_api_metadata_local_check

* Adding protocol and port option to local nova metadata api check.


tonytan4ever:

PR #478, opened 2018-03-27
TURTLES-713 Properly configure agent.plugin timeout value

npawelek:
Properly configure agent.plugin timeout value for all plugin templates.

### DIFF
--- a/playbooks/files/rax-maas/plugins/swift-recon.py
+++ b/playbooks/files/rax-maas/plugins/swift-recon.py
@@ -432,6 +432,9 @@ def make_parser():
     parser.add_argument('--swift-recon-path',
                         default='/usr/local/bin',
                         help='The path for the swift-recon directory.')
+    parser.add_argument('-t',
+                        default='30',
+                        help='Set the swift-recon timeout.')
     return parser
 
 
--- a/playbooks/files/rax-maas/plugins/swift-recon.py
+++ b/playbooks/files/rax-maas/plugins/swift-recon.py
@@ -434,7 +434,7 @@ def make_parser():
                         help='The path for the swift-recon directory.')
     parser.add_argument('-t',
                         default='30',
-                        help='Set the swift-recon timeout.')
+                        help='Set a timeout value in seconds for swift-recon.')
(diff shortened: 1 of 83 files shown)


playbooks/files/rax-maas/plugins/swift-recon.py
```diff
@@ -432,6 +432,9 @@ def make_parser():
     parser.add_argument('--swift-recon-path',
                         default='/usr/local/bin',
                         help='The path for the swift-recon directory.')
+    parser.add_argument('-t',
+                        default='30',
```
BjoernT: Please spell out the units like seconds or milli seconds etc.

playbooks/templates/rax-maas/ceph_cluster_stats.yaml.j2
```diff
@@ -16,6 +16,7 @@ disabled    : "{{ (inventory_hostname != groups['mons'][0] or check_name | match
 details     :
     file    : run_plugin_in_venv.sh
     args    : {{ ceph_args }}
+    timeout : {{ (maas_check_timeout_override[label] | default(maas_check_timeout) * 1000) }}
```
BjoernT: Lets not change all templates and much rather calculate it via a default variable and convert maas_check_timeout into maas_check_timeout * 1000 if we detects the input is seconds as example (ie the timeout is only 3 digits long etc)

npawelek: The problem is that the check-level still utilizes seconds, but doesn't actually do anything for `agent.plugin` checks. These check types require a millisecond declaration:

https://developer.rackspace.com/docs/rackspace-monitoring/v1/tech-ref-info/check-type-reference/#agent-plugin

playbooks/files/rax-maas/plugins/swift-recon.py
npawelek: done

npawelek:

npawelek:

npawelek:

PR #482, opened 2018-03-29
Lowering default monitoring thresholds for rabbit

BjoernT:
As the result of PR476 we increased default values for
rabbitMQ in case queues filling up (when without consumers
as example) to particular high values. OPS engineers should
know sooner when issues arrive in order to prevent the rabbitMQ
issues from growing before they become to large manage.
Therefore new default values are proposed which can be
overridden if necessary :

- maas_rabbitmq_queued_messages_excluding_notifications_threshold: 1000
- maas_rabbitmq_messages_without_consumers_threshold: 5000

### DIFF
--- a/playbooks/templates/rax-maas/nova_console_check.yaml.j2
+++ b/playbooks/templates/rax-maas/nova_console_check.yaml.j2
@@ -1,12 +1,12 @@
 {% from "templates/common/macros.jinja" import get_metadata with context %}
 {% set label = "nova_console_check" %}
 {% set check_name = label+'--'+inventory_hostname %}
-{% if nova_console_type == 'spice' %}
+{% if maas_nova_console_type == 'spice' %}
 {% set console_service_name = 'nova_spice' %}
-{% set nova_console_base_url = 'spice_auto.html' %}
-{% elif nova_console_type == 'novnc' %}
+{% set maas_nova_console_base_url = 'spice_auto.html' %}
+{% elif maas_nova_console_type == 'novnc' %}
 {% set console_service_name = 'nova_novnc' %}
-{% set nova_console_base_url = 'vnc_auto.html' %}
+{% set maas_nova_console_base_url = 'vnc_auto.html' %}
 {% endif %}
 type        : agent.plugin
 label       : "{{ check_name }}"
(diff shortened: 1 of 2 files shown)


npawelek:
@BjoernT @jedsmith Looking at the data around these, they roughly equate to 5% of the overall alarm volume (excluding the inflated network throughput alarms which skews data). This is is still somewhat significant. 

The trifecta of alarms that we reworked fired off the following times:
```
rabbitmq_queues_without_consumers 1215
rabbitmq_qgrowth_excl_notifications   873
rabbitmq_msgs_excl_notifications        1166
```

Of these `3254` alarms total across these three alarms, `2658` alarms were successfully sent to the npManaged notification plan (across all rabbit alarms). With that in mind, theres a discrepancy meaning some of those numbers were also caused from plugin failures.
```
Check Name                                       Exit Failure     Successful Alarm
rabbitmq_status                                          1694                 2658
```

Ultimately, we need to have a happy medium for these thresholds. While I feel they're arbitrary to an extent, they're not overboard. We're already running the `5000` value for GE and have been for some time. When you think about the time differentials between an alarm firing off and someone actually working it, they usually resolve themselves and simply continually flap with ticket updates. This is what I'm trying to avoid, and insulating with a higher number seemed to be the best option. When there is an issue, it will most likely be valid. If we lower unconsumed queues to 5000, we're going to get false positives from things such as computes going down in an environment.

To properly rework these thresholds I think we need to truly understand the metrics from the notifications of these alarms. This information is contained in hive, but we didn't collect it in the rush to gather data. Potentially this may be an avenue we can use. @CrashenX Would that be feasible?

BjoernT:
Yes it is true that a message growth occurs if a compute node is down but saying 5000 will lead to false alerts vs 20000 can't simply be sustained. The likelihood is even that limit will be hit and BTW it is NOT a false alert. If the compute node is down for too long we need to remove the queue, it's a legitimate problem.

npawelek:
@BjoernT If that's the case then yes, I agree lowering the value is valid. However, this information would need to become doctrine and disseminated to the engineering team. Currently, there's a lot of questionability around deleting queues, especially for nodes that are down due to hardware issues with messages backed up. You also need to consider that MaaS **is not** good mechanism for "lets wait and take care of this later", so at the time of which an alarm fires off in this context... there should be a specific set of guidelines that needs to be followed. My issue with the low values is simply that the issue either disappears before being worked or becomes spammy and subsequently glanced over, ultimately causing more workload.

npawelek:
@BjoernT I've asked that we gather additional metric data around this issue so that we can tune thresholds with non-arbitrary values. 

cc: @rebeccaskinner, @CrashenX

jedsmith:

jedsmith:
DNMing pending data collection efforts. Let's hold off on this until we get toward a resolution.

BjoernT:
Closing for now pending investigation

PR #493, opened 2018-04-17
TURTLES-761: Read maas_rally scenario config from file

cfarquhar:
This change eases troubleshooting burden by moving scenario configuration items
from the plugin CLI arguments to a configuration file.  This allows operators
to simply run the plugin without first looking up the extra arguments from MaaS
agent configuration files.

- Renamed `extra_vars` dictionary to `task_args`
- Move `times` and `concurrency` variables into the new `task_args` dictionary
- Updated playbooks to account for `times` and `concurrency` moving into
  `task_args`
- Update the plugin to read task arguments (e.g. `times`, `concurrency`, and
  items previously stored in `extra_vars` like network UUIDS) from the config
  file instead of from CLI arguments.


cfarquhar:

PR #495, opened 2018-04-18
JIRA: REO-218

shannonmitchell:
Fix missing modules in venv

During our upgrades in phobos we found that the script wrapper
was pointing to an old venv.  After upgrading it to the new
maas-1.7.0 venv, we flooded the queue with tickets due to missing
venv modules.  I pulled all of the missing values from
/tmp/pip-constraints.txt and updated the vars/maas-agent.yml to
include them.  After re-running the site.yml all alearts cleared.


shannonmitchell:
May be the wrong approach, as it looks like these packages should be installed in some of the many playbooks spread out for each set of services.

PR #491, opened 2018-04-17
Fixing maas gate failure.

tonytan4ever:
1.  Pin pip to upstream restrictions.
2.  Add `no_log: True` to comply OSA linting.
3. Pin SecretStorage to Python 2 compatible version. see details [here](https://github.com/mitya57/secretstorage).

PR #496, opened 2018-04-19
Pin pip packages for Ceph installation

andymcc:


tonytan4ever:
This has been incorporated into https://github.com/rcbops/rpc-maas/pull/491

PR #499, opened 2018-04-24
TURTLES-762

tonytan4ever:
Add common notification metadata section

### DIFF
--- a/playbooks/templates/rax-maas/ceph_rgw_stats.yaml.j2
+++ b/playbooks/templates/rax-maas/ceph_rgw_stats.yaml.j2
@@ -1,3 +1,4 @@
+{% from "templates/common/macros.jinja" import get_metadata with context %}
 {% set label = "ceph_rgw_stats" %}
 {% set check_name = label+'--'+inventory_hostname %}
 {% set ceph_args = [maas_plugin_dir + "/ceph_monitoring.py", "--name", "client.raxmon", "--keyring", "/etc/ceph/ceph.client.raxmon.keyring"] %}
@@ -14,6 +15,8 @@ details     :
     file    : run_plugin_in_venv.sh
     args    : {{ ceph_args }}
     timeout : {{ (maas_check_timeout_override[label] | default(maas_check_timeout) * 1000) }}
+{{ get_metadata(label).strip() }}
+{# Add extra metadata options with two leading white spaces #}
 alarms      :
     ceph_rgw_status :
         label                   : ceph_rgw_status.{{ inventory_hostname }}
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,5 @@ tests/*.retry
 # Eclipse artifacts
 .project
 .pydevproject
+openstack-ansible-ops
+*.retry
--- a/playbooks/maas-pre-flight.yml
+++ b/playbooks/maas-pre-flight.yml
@@ -34,3 +34,25 @@
     maas_pw_check: "{{ lookup('file', playbook_dir + '/../tests/user_rpcm_secrets.yml') | from_yaml }}"
   tags:
     - maas-pre-flight
+
+
+- name: Check metadata variables are defined
+  hosts: localhost
+  connection: local
+  gather_facts: false
+  tasks:
+    - name: Get rendered test metadata template string
+      set_fact:
+        rendered_maas_metata_test_tmpl: "{{ lookup('template', playbook_dir + '/templates/rax-maas/test_metadata.yaml.j2') }}"
+
+    - name: Check metadata variables are defined
+      fail:
+        msg: >-
+          There are missing metadata environment variable(s), as you can see 'unknown' from
+          this test template '{{ rendered_maas_metata_test_tmpl }}'. Please set proper env
+          variables, possible options are: maas_env_identifier, maas_rpco_dir, maas_osa_dir
+      when:
+        - maas_pre_flight_metadata_check_enabled | default(true) | bool
+        - rendered_maas_metata_test_tmpl | search('unknown')
+  tags:
+    - maas-pre-flight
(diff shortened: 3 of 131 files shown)


playbooks/templates/common/metadata_common.yaml
```diff
@@ -0,0 +1,10 @@
+metadata          :
+  RPC: {{ maas_rpc | default(true) }}
+  RPC_env_identifier: {{ maas_env_identifier | default('unkown')}}
+  RPC_env_identifier_alt: {{ maas_env_identifier_alt | default('unkown')}}
+  RPC_maas_version: {{  lookup('pipe', 'cd .. && git describe --tags --abbrev=0') }}
+  RPC_deploy_date: {{ ansible_date_time.date }}
+  RPC_check_category: {{ check_category | default('Unkown') }}
+  RPC_datacenter: {{ maas_rpc_datacenter | default("CDC") }}
+  RPC_device_number: {{ get_server_numbr | default('unkown')}}
```
npawelek: typos:
`get_server_numbr`
`unkown` (this is in multiple places)

playbooks/templates/common/metadata_common.yaml
```diff
@@ -0,0 +1,10 @@
+metadata          :
+  RPC: {{ maas_rpc | default(true) }}
+  RPC_env_identifier: {{ maas_env_identifier | default('unkown')}}
+  RPC_env_identifier_alt: {{ maas_env_identifier_alt | default('unkown')}}
+  RPC_maas_version: {{  lookup('pipe', 'cd .. && git describe --tags --abbrev=0') }}
+  RPC_deploy_date: {{ ansible_date_time.date }}
+  RPC_check_category: {{ check_category | default('Unkown') }}
```
npawelek: I'm not seeing where this is set. It would have to be defined in each template, correct? We probably want to prepend `maas_` to the variable name.

playbooks/templates/common/metadata_common.yaml
```diff
@@ -0,0 +1,10 @@
+metadata          :
+  RPC: {{ maas_rpc | default(true) }}
+  RPC_env_identifier: {{ maas_env_identifier | default('unkown')}}
+  RPC_env_identifier_alt: {{ maas_env_identifier_alt | default('unkown')}}
+  RPC_maas_version: {{  lookup('pipe', 'cd .. && git describe --tags --abbrev=0') }}
```
npawelek: This is a relative path, will this work in all use-cases?

playbooks/templates/common/metadata_common.yaml
```diff
@@ -0,0 +1,10 @@
...
+  RPC_deploy_date: {{ ansible_date_time.date }}
```
npawelek: As playbooks are iterative, if we used a more specific timestamp such as epoch, would each template receive a different timestamp or would it be based on the initial execution of the playbook?

playbooks/templates/common/metadata_common.yaml
npawelek: Also, how are we defining this per-host? Can we leverage raxmon to strip it off the entityURI?

tonytan4ever: I didn't quite yet figure out how to get the server number yet,  we will need to chat about this.

We can utilize inventory_hostname and use raxmon entity URL for sure.

playbooks/templates/common/metadata_common.yaml
tonytan4ever: Done.

playbooks/templates/common/metadata_common.yaml
tonytan4ever: Based on my test it will read from the playbook directory so it will work.

playbooks/templates/common/metadata_common.yaml
tonytan4ever: It will be based on the initial execution of the playbook.

playbooks/templates/rax-maas/ceph_rgw_stats.yaml.j2
```diff
@@ -9,11 +9,14 @@ type        : agent.plugin
 label       : "{{ check_name }}"
 period      : "{{ maas_check_period_override[label] | default(maas_check_period) }}"
 timeout     : "{{ maas_check_timeout_override[label] | default(maas_check_timeout) }}"
-disabled    : "{{ (check_name | match(maas_excluded_checks_regex)) | ternary('true', 'false') }}"
+disabled    : "{{ (inventory_hostname != groups['rgws'][0] or check_name | match(maas_excluded_checks_regex)) | ternary('true', 'false') }}"
```
npawelek: @tonytan4ever can you rebase? Looks like this was reverted: https://github.com/rcbops/rpc-maas/pull/502

tonytan4ever:

npawelek:

tonytan4ever:

andymcc:
If rpc-maas requires a version of Ansible that isn't in line with OSA/ceph-ansible or others (including the jinja2 versions) rpc-maas should probably pin it's own ansible/versions in it's own venv or something similar.

This change suggests I won't be able to run rpc-maas against a ceph deployment unless I then create a version of ansible somewhere that is not using jinja2 > 2.8, which would not be true of any ceph or OSA deploys (pike+)

Alternatively the approach needs a rethink to work with the newer versions of jinja2.

This problem I could see is that this creates a workflow like this, for both ceph and pike+ openstack deploys:

Bootstrap ansible for Ceph/Openstack (this will deploy ansible and jinja2 at least 2.9.6 for pike, 2.10 in queens+)
Run openstack/ceph deploy
Run MaaS playbooks - this fails now.
Deploy ansible and jinja2 somewhere else (I wouldn't want to deploy an older version of jinja2 just for the purpose of MaaS).
Run MaaS playbooks using the newly deployed ansible /w jinja2==2.8.

I think that will create issues in the long run, and isn't particularly friendly for prod deployments or deployments in general.

tonytan4ever:

tonytan4ever:
recheck_cit_container_ansible-lint_tox-test

tonytan4ever:
@andymcc:  Adjusted macro and removed hard dependency of Jinja2==2.8

PR #500, opened 2018-04-25
TURTLES-769: Add alternate logic in hp_monitoring.py for encrypted volumes

npawelek:
Properly validate logical volume status if HP volume is encrypted.


npawelek:
recheck_all

PR #502, opened 2018-04-30
Revert limiting enablement of rgw checks to first node of group

npawelek:
Revert limiting enablement of rgw checks to first node of group. This was an incorrect assumption as part of the NVA reduction work.


npawelek:

PR #508, opened 2018-05-09
Fix non-quoted metric in Octavia quota template

npawelek:
Fix non-quoted metric in Octavia quota template


tonytan4ever:

tonytan4ever:

PR #511, opened 2018-05-17
Making metadata working with ceph deployments

tonytan4ever:


tonytan4ever:

tonytan4ever:
@npawelek :  for situations like comcast this morning we will need to overwrite _maas_env_product_  and provide corresponding _maas_product_dir_ and  _maas_product_osa_dir_. These two can be the same and have to be the place OSA is cloned/downloaded.

I think this way the logic is consistent with current code and its behavior.

npawelek:
@tonytan4ever In this case, the osa dir will become the product dir. There should be no need for having the osa dir when osa is the set product.

PR #513, opened 2018-05-21
Upgrade virtualenv creation method and fixing gating

tonytan4ever:


tonytan4ever:
recheck_xenial_pike_deploy
recheck_trusty_liberty_deploy
recheck_trusty_mitaka_deploy

tests/aio-create.sh
```diff
@@ -145,6 +145,7 @@ if [ "${RE_JOB_SCENARIO}" == "ceph" ]; then
   git clone https://github.com/rcbops/rpc-ceph /opt/rpc-ceph
   export RPC_MAAS_DIR="$(pwd)"
   pushd /opt/rpc-ceph
+    apt-get install -y apt-transport-https
```
mattt416: Hey @tonytan4ever  It's not clear to me why this package is only installed on ceph/newton -- can you add some details to commit message to clarify?

test-requirements.txt
```diff
@@ -5,7 +5,7 @@ bashate>=0.2 # Apache-2.0
 flake8<2.6.0,>=2.5.4 # MIT
 pyasn1 # BSD
 pyOpenSSL>=0.14 # Apache-2.0
-requests!=2.12.2,!=2.13.0,>=2.10.0 # Apache-2.0
+requests[security]
```
mattt416: @tonytan4ever I don't see us installing this package anywhere else, can you document in the commit message why this is needed? Also, what is the implication of removing the pin for requests that was previously there?

tests/aio-create.sh
```diff
@@ -214,15 +220,26 @@ pushd /opt/openstack-ansible
   # Disable tempest on older releases
   sed -i '/.*run-tempest.sh.*/d' scripts/gate-check-commit.sh  # Disable the tempest run
 
-  # NOTE(tonytan4ever): pin pip to the last working version before pip 10
+  # NOTE(tonytan4ever): pin pip to the last working version before pip 10
```
mattt416: This change seems to be only adding whitespace?

tests/aio-create.sh
```diff
@@ -214,15 +220,26 @@ pushd /opt/openstack-ansible
...
   sed -i '/.*run-tempest.sh.*/d' scripts/gate-check-commit.sh  # Disable the tempest run
 
-  # NOTE(tonytan4ever): pin pip to the last working version before pip 10
+  # NOTE(tonytan4ever): pin pip to the last working version before pip 10	
   pip install -U -r /opt/openstack-ansible/global-requirement-pins.txt
 
   # Pin python-ldap so it stops breaking everything
+  # Fixing missing netbase packages
   echo "python-ldap<3;python_version=='2.7'" >> /opt/openstack-ansible/global-requirement-pins.txt
+  sudo apt-get -o Dpkg::Options::="--force-confmiss" install --reinstall netbase
```
mattt416: Can you clarify why this is needed?

tests/test-ansible-functional.sh
```diff
@@ -154,7 +154,9 @@ function enable_maas_api {
   #                  different virtualenv which are done in three tasks. These three tasks are being
   #                  run in this particular order to ensure "pyrax" installs correctly.
   PYTHON_BIN="$(which python)"
-  virtualenv --python="${PYTHON_BIN}" /opt/test-maas
+  virtualenv --python="${PYTHON_BIN}" --no-pip /opt/test-maas
+  curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
```
mattt416: I don't see any issue w/ this change here, but curious why it's needed. Got more info?

tests/aio-create.sh
tonytan4ever: This is because in newton we will encounter this error if this is not installed:

```
PLAY [Deploy Grafana] **********************************************************

TASK [Ensure https repos function] *********************************************
Tuesday 22 May 2018  19:15:33 +0000 (0:00:00.226)       0:16:14.559 *********** 
ok: [aio1]

TASK [Add grafana apt-keys] ****************************************************
Tuesday 22 May 2018  19:15:35 +0000 (0:00:02.031)       0:16:16.591 *********** 
changed: [aio1]

TASK [Add grafana repo] ********************************************************
Tuesday 22 May 2018  19:15:37 +0000 (0:00:02.429)       0:16:19.020 *********** 
changed: [aio1]

TASK [Update Apt cache] ********************************************************
Tuesday 22 May 2018  19:15:38 +0000 (0:00:00.802)       0:16:19.823 *********** 
FAILED - RETRYING: TASK: Update Apt cache (5 retries left).
FAILED - RETRYING: TASK: Update Apt cache (4 retries left).
FAILED - RETRYING: TASK: Update Apt cache (3 retries left).
FAILED - RETRYING: TASK: Update Apt cache (2 retries left).
FAILED - RETRYING: TASK: Update Apt cache (1 retries left).
fatal: [aio1]: FAILED! => {"changed": false, "failed": true, "module_stderr": "", "module_stdout": "", "msg": "MODULE FAILURE"}
```

and this in ceph:
```
TASK [Update Apt cache] ********************************************************
Tuesday 22 May 2018  04:02:00 +0000 (0:00:01.211)       0:19:41.006 *********** 
FAILED - RETRYING: Update Apt cache (5 retries left).
changed: [mon1]
changed: [mon2]
changed: [osd2]
changed: [osd1]
FAILED - RETRYING: Update Apt cache (4 retries left).
changed: [rsyslog1]
changed: [allsvc]
changed: [rgw1]
FAILED - RETRYING: Update Apt cache (3 retries left).
FAILED - RETRYING: Update Apt cache (2 retries left).
FAILED - RETRYING: Update Apt cache (1 retries left).
fatal: [localhost]: FAILED! => {"attempts": 5, "changed": false, "failed": true, "msg": "Failed to update apt cache."}
```
(This is when installing telegraf)

@andymcc's advice aded apt-transport-https to fix them

mattt416: @tonytan4ever thanks! So what happens if we deploy this code for a customer on newton? This is only getting installed in `tests/aio-create.sh` but the listed failure appears to be an installation-related task.

mattt416: @tonytan4ever ok, tracked down @andymcc's commit message in rpc-ceph and think I understand this better now. Perhaps it'd be better to just ensure this package is installed across the board [here](https://github.com/rcbops/rpc-maas/blob/master/tests/aio-create.sh#L137)?

test-requirements.txt
tonytan4ever: This is for fixing the error:

```UPPER_CONSTRAINTS_PROTO=http
++ export 'UPPER_CONSTRAINTS_FILE=https://git.openstack.org/cgit/openstack/requirements/plain/upper-constraints.txt?h=master'
++ UPPER_CONSTRAINTS_FILE='https://git.openstack.org/cgit/openstack/requirements/plain/upper-constraints.txt?h=master'
++ pip install --requirement requirements.txt --constraint 'https://git.openstack.org/cgit/openstack/requirements/plain/upper-constraints.txt?h=master'
/usr/local/lib/python2.7/dist-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:318: SNIMissingWarning: An HTTPS request has been made, but the SNI (Subject Name Indication) extension to TLS is not available on this platform. This may cause the server to present an incorrect TLS certificate, which can cause validation failures. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#snimissingwarning.
  SNIMissingWarning
/usr/local/lib/python2.7/dist-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:122: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
Certificate did not match expected hostname: git.openstack.org. Certificate: {'subject': ((('organizationalUnitName', u'Domain Control Validated'),), (('organizationalUnitName', u'PositiveSSL'),), (('commonName', u'git.airshipit.org'),)), 'subjectAltName': (('DNS', 'git.airshipit.org'), ('DNS', 'www.git.airshipit.org')), 'notAfter': 'May 19 23:59:59 2019 GMT'}
Exception:```

File: tests/aio-create.sh (hunk 3 above)
Comment (tonytan4ever) [reply tests/aio-create.sh]: Oops, this can be gone.

File: tests/aio-create.sh (hunk 4 above)
Comment (tonytan4ever) [reply tests/aio-create.sh]: This is for this error from _Add in cinder devices types_  in _os_cinder_ role:

```
File \"/openstack/venvs/cinder-16.0.14/lib/python2.7/site-packages/eventlet/support/dns/rdtypes/IN/WKS.py\", line 23, in <module>", "    _proto_tcp = socket.getprotobyname('tcp')", "socket.error: protocol not found"], "stdout": "", "stdout_lines": []
```

This is also per @andymcc's advice: https://rackspace.slack.com/archives/C8TEU9GUT/p1526996313000235

File: tests/test-ansible-functional.sh (hunk 5 above)
Comment (tonytan4ever) [reply tests/test-ansible-functional.sh]: For some reason the pip inside virtualenv is causing error:

```unning virtualenv with interpreter /var/lib/jenkins/workspace/PR_rpc-maas-master-trusty-mitaka-deploy/rpc-maas/.tox/functional/bin/python
Using real prefix '/usr'
New python executable in /opt/test-maas/bin/python
Installing setuptools, pip, wheel...done.
  File "/opt/test-maas/bin/pip", line 7, in <module>
    from pip._internal import main
ImportError: No module named _internal```
```

We tried pretty hard but it's hard to reproduce.
By going over the internet seems like installing a fresh pip inside virutalenv would fix it, as it did.

test-requirements.txt
mattt416: `UPPER_CONSTRAINTS_FILE` points to http:// and not https:// as per your snippet, which I believe is what fixed this issue. I'd double-check if this is actually needed.

tests/test-ansible-functional.sh
mattt416: OK -- I remember hitting that error when we were testing stuff out manually. I can't remember which version of pip was automatically being installed in the venv though, I'd need to check my chat history.

tests/aio-create.sh
mattt416: @tonytan4ever OK, understood. Now can you not just install it as per normal (as it doesn't appear to be in the nodepool images by default) without doing all this install --reinstall business? We may want to check w/ @odyssey4me if that package should be installed by default, as it does seem sort of critical.

tests/aio-create.sh
```diff
@@ -219,10 +223,21 @@ pushd /opt/openstack-ansible
...
   # Setup an AIO
   sudo -H --preserve-env ./scripts/gate-check-commit.sh
+
+  # # NOTE(tonytan4ever): Help debugging the apt issue.
+  # Implement debug output for apt so that we can see more information
+  # about whether the 'Acquire-by-hash' feature is being used, and what
+  # might be causing it to fall back to the old style.
+  # This config file should be copied into containers by the lxc_hosts
+  # role.
+  # ansible hosts -m lineinfile -a 'create=yes dest=/etc/apt/apt.conf.d/99debug line="Debug::Acquire::http \"true\";"'
```
odyssey4me: This is commented, so it does nothing. Also it'd be more useful to have the debug output in the AIO during the initial build, not after it's built - so this would be better to do, before gate-check-commit executes: https://github.com/rcbops/rpc-openstack/commit/b0215bbf69c9b6eaf148e332a32564a0101c88de#diff-93518fda10b0403d3c5c20b4df4740a6

tests/test-ansible-functional.sh
odyssey4me: get-pip.py may already exist on the host - add a check if it's already in /opt/get-pip.py before downloading it again, otherwise relying on an external resource may cause a failure

Also, there's a robust function for this in https://github.com/openstack/openstack-ansible/blob/master/scripts/scripts-library.sh#L355-L381 which you could either re-use, or copy and use. It does retries and also executes against a primary and fallback URL. You can also use that function to get the right version of pip/setuptools/wheel into the venv you're building here, something like this: https://github.com/openstack/openstack-ansible/blob/master/scripts/bootstrap-ansible.sh#L154-L160

tests/test-ansible-functional.sh
```diff
@@ -154,7 +154,9 @@ function enable_maas_api {
...
+  /opt/test-maas/bin/python get-pip.py
   /opt/test-maas/bin/pip install "jinja2" --isolated --upgrade --force-reinstall
```
odyssey4me: All these requirements being added into the venv are unconstrained. It may be best to build a working venv and use 'pip freeze' to construct a constraint file, then use the dependency update hook to propose weekly updates to the constraints file to keep them current.

With a constraints file you'd have less failures from external factors hitting the repo and would be able to manage their effects on important work being done.

tox.ini
```diff
@@ -36,7 +37,10 @@ setenv =
 
 [testenv:bindep]
 # Deps is empty on purpose
-deps =
+install_command =
```
odyssey4me: These inherit from testenv, so they're unnecessary additions.

tox.ini
```diff
@@ -36,7 +37,10 @@ setenv =
...
+    pip install {opts} {packages} --isolated
+deps =
```
odyssey4me: These inherit from testenv, so they're unnecessary additions.

tox.ini
```diff
@@ -132,6 +136,7 @@ install_command =
     pip install {opts} {packages} --isolated --upgrade --force-reinstall
```
odyssey4me: Are constraints not used here on purpose? Why not just inherit this command from the testenv like all the others?

tox.ini
```diff
@@ -132,6 +136,7 @@ install_command =
...
 deps =
     -r{toxinidir}/test-requirements.txt
+	requests[security]
```
odyssey4me: testenv already has these, why not remove this and let it inherit from there?

tests/aio-create.sh
```diff
@@ -182,13 +182,17 @@ pushd /opt/openstack-ansible
...
+    export UPPER_CONSTRAINTS_FILE="http://git.openstack.org/cgit/openstack/requirements/plain/upper-constraints.txt?id=$(awk '/requirements_git_install_branch:/ {print $2}' playbooks/defaults/repo_packages/openstack_services.yml) -U"
 
   elif [ "${RE_JOB_SCENARIO}" == "newton" ]; then
     git checkout "stable/newton"  # Branch checkout of Newton (Current Stable)
     enable_ironic
     # NOTE(tonytan4ever): newton needs this to get around gating:
     # https://rackspace.slack.com/archives/CAD5VFMHU/p1525445460000172
     add_lxc_overrides
+    if [ "${RE_JOB_IMAGE}" == "trusty" ]; then
+      export UPPER_CONSTRAINTS_FILE="http://git.openstack.org/cgit/openstack/requirements/plain/upper-constraints.txt?id=$(awk '/requirements_git_install_branch:/ {print $2}' playbooks/defaults/repo_packages/openstack_services.yml) -U"
```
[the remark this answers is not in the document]
tonytan4ever: Well,  this is because trusty-newton breaks and xenial-newton doesn't

tonytan4ever:
recheck_all

tests/aio-create.sh
```diff
@@ -219,10 +223,21 @@ pushd /opt/openstack-ansible
 
   # Pin python-ldap so it stops breaking everything
   echo "python-ldap<3;python_version=='2.7'" >> /opt/openstack-ansible/global-requirement-pins.txt
+  # Fixing missing netbase packages
+  sudo apt-get -o Dpkg::Options::="--force-confmiss" install --reinstall netbase
```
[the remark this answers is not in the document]
tonytan4ever: Moved to line 137

tox.ini
tonytan4ever: I think so.

tox.ini
tonytan4ever: done

PR #516, opened 2018-06-01
Remove conflicting nova_console_type override

BjoernT:
`nova_console_type` is renamed to `maas_nova_console_type`
along with `nova_console_ports` which is also renamed to
`maas_nova_console_ports` in order to prevent variable
inheritance issues deriving from OSA use.


npawelek:
@BjoernT Please add this when you get a chance

BjoernT:
@tonytan4ever Notes added, strangely i never got your update

PR #518, opened 2018-06-05
RE-1483 Use updated gating hook dirs

mattt416:
We are moving from `gating/pre_merge_test` to `gating/check` and
`gating/post_merge_check` to `gating/periodic`. This commit updates
the gating hook dirs accordingly. This same exercise has been
completed for rpc-ceph, so making this adjustment in
tests/aio-create.sh is also safe to make.

Lastly, we create symlinks for `gating/gate` and `gating/release`,
both of which point to `gating/check`. This can be adjusted by the
team at a later date.


mattt416:
Hi @tonytan4ever , can you please review this when you get a moment? We need these changes to go in before the end of the month so we can release rpc-maas via pipeline on release day.

Thanks!

PR #520, opened 2018-06-06
Update maas_checks_util for Python 2, new checks

briancurtin:
Per INTAKERPC-120 and TURTLES-802, this utility needs to be changed to work on Python 2.7 as well as updated to handle newly added checks. Additionally, there was an ask to utilize the categorization that check templates now expose.

This change updates the Jinja undefined handler to work with both Python 2 and Python 3 as well as the necessary changes to stop ignoring some newer checks that we were skipping.

Additionally, the module was renamed slightly so it's easier to talk about both the docs end of this and the rpc-maas end of it.

### DIFF
--- a/doc/utils/README.rst
+++ b/doc/utils/README.rst
@@ -1,9 +1,13 @@
-maas_checks_util.py
-===================
+maas_checks_config.py
+=====================
 
-The ``maas_checks_util`` module includes a ``check_details`` generator
-which yields check names and corresponding details about them. These
-include the alarms within each check as well as the criteria that
+The ``maas_checks_config`` module includes a ``check_details`` generator
+which yields check names and corresponding details about them,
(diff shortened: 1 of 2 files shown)


tonytan4ever:
recheck_trusty_liberty_deploy
recheck_trusty_mitaka_deploy
recheck_trusty_newton_deploy
recheck_xenial_newton_deploy

briancurtin:
I would just ignore the gates as this change is purely for a documentation tool, and any failure will be totally unrelated. As long as it gets code review approval we can merge it.

mattt416:
@briancurtin @tonytan4ever I created https://github.com/rcbops/rpc-gating/pull/872 to skip the functional tests when there's a change to `doc/`, etc. Can you two have a peek when you get a moment?

PR #522, opened 2018-06-14
CEPHSTORA-304 Allow MaaS to use a different radosgw_address

andymcc:
This PR allows you to use an address for Rados GW that is not the
default ansilbe_host address.

The method utilises the existing ceph-ansible variables to define the
radosgw_address, radosgw_address_block or radosgw_interface, as you
would in ceph-ansible. This is to ensure that ceph-ansible has
configured Rados GW to listen on the same address as we set up MaaS to
monitor against.


tonytan4ever:
LGTM

PR #523, opened 2018-06-19
TURTLES-894: Add a new way of checking neutron agents status

tonytan4ever:


tonytan4ever:
Close as per latest discussion with @npawelek.

PR #524, opened 2018-06-19
Move check configuration docs into this repo

briancurtin:
MaaS configuration docs used to be handled within privatecloud-docs by doing a checkout of this repo, importing maas_checks_config out of a temporary directory, and rendering Sphinx tables from that output. After meeting with several docs folks it was suggested to have the docs live fully within this repo and be served via Deconst out of here. This reduces the complexity of building privatecloud-docs as it doesn't depend on an external source and allows this documentation to change on its own.

This introduces per-category docs via the index, and a single-page option linked at the top.

Additionally, this introduces a separate docs build from some existing docs. This all stems from an ask to get configuration info into customers and account managers' hands more easily, and this info is specific to something that already existed elsewhere, which is why it's currently kept separate from any other docs in this project.


tonytan4ever:
recheck_trusty_liberty_deploy
recheck_xenial_ceph_deploy

tonytan4ever:
Ceph gate is having some issues, this PR will neither impact nor be impacted by ceph gate, so merging.

PR #527, opened 2018-06-21
Remove the filebeat check

cloudnull:
The filebeat checks are in conflict with the direction we're taking the
rest of the Elastic stack. The Elastic stack work is now running running
beats (not just filebeat) on hosts machines and is limiting beat
installations within containers. This is on top of the fact that the
RAX-MaaS checks for filebeat are largely meaningless and causes cascading
alerts/tickets especially when any one process is disabled or otherwise
removed. Log shipping to the logging host is taken care of with rsyslog,
and journald (journal-remote). While filebeat is useful, it's only an
analytic tool and not something that is mission critical requiring an
alert if it stops for arbitrary reasons.


npawelek:
@cloudnull Works for me. To provide some context, this check has only been useful in discovering gaps in the deployment process. It seems that filebeat is more often then not left out of node additions and QC's. It might be something to think about regarding the new architecture as there will be no replacement check to perform validation.

tonytan4ever:
recheck_trusty_newton_deploy
recheck_xenial_newton_deploy

tests/user_liberty_vars.yml
```diff
@@ -26,10 +26,6 @@ maas_excluded_alarms:
   - '^Network transmit rate.*'
   - '^percentage_disk_utilisation.*'
 
-# swift-recon does not have '--time' option in Liberty
-maas_excluded_checks:
```
tonytan4ever: This may introduced a gate error. See comment below

tonytan4ever: TASK [Install ironic ironic-compute checks] ************************************
Thursday 21 June 2018  15:43:07 +0000 (0:00:00.804)       0:03:04.504 ********* 
fatal: [aio1_ironic_compute_container-f92f0cbb -> 172.29.236.100]: FAILED! => {"changed": false, "failed": true, "msg": "AnsibleUndefinedVariable: 'maas_excluded_checks_regex' is undefined"}

cloudnull: @tonytan4ever I'm not sure how the removal of that var is impacting newton ?

cloudnull:
Abandoned in favor of the combined commit https://github.com/rcbops/rpc-maas/pull/528

PR #529, opened 2018-06-22
TURTLES-894: WARN neutron/nova agents/computes/schedulers node if can't reach API

tonytan4ever:
@npawelek:   As per our conversation,  when the API node can't be reached from agent/computer nodes: https://github.com/rcbops/rpc-maas/blob/master/playbooks/files/rax-maas/plugins/neutron_service_check.py#L35, We generate a warning instead of letting plugin exit(1)

### DIFF
--- a/playbooks/files/rax-maas/plugins/neutron_service_check.py
+++ b/playbooks/files/rax-maas/plugins/neutron_service_check.py
@@ -15,11 +15,13 @@
 # limitations under the License.
 
 import argparse
+import sys
 
 from maas_common import get_neutron_client
 from maas_common import metric_bool
 from maas_common import print_output
 from maas_common import status_err
+from maas_common import status_err_no_exit
 from maas_common import status_ok
 
 
@@ -33,7 +35,8 @@ def check(args):
     # not gathering api status metric here so catch any exception
     except Exception as e:
         metric_bool('client_success', False, m_name='maas_neutron')
-        status_err(str(e), m_name='maas_neutron')
+        status_err_no_exit(str(e), m_name='maas_neutron')
+        sys.exit(0)
     else:
         metric_bool('client_success', True, m_name='maas_neutron')
 
--- a/playbooks/files/rax-maas/plugins/nova_service_check.py
+++ b/playbooks/files/rax-maas/plugins/nova_service_check.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import argparse
+import sys
 
 from maas_common import get_auth_ref
 from maas_common import get_keystone_client
@@ -22,6 +23,7 @@ from maas_common import get_nova_client
 from maas_common import metric_bool
 from maas_common import print_output
 from maas_common import status_err
+from maas_common import status_err_no_exit
 from maas_common import status_ok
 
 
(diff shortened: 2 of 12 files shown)


npawelek:
@tonytan4ever This addition won't change the behavior or what we're dealing with, rather it adds an additional criteria to check for while leaving the existing issue unaltered.

The intent that we decided on was to apply different logic to the same metric we're gathering around `neutron-dhcp-agent_status`, `nova-compute_status`, and others affected by APIs being down. We likely need to manipulate the existing metrics we're creating. For example, the plugin `nova_service_check.py` used by `nova_compute_check` should have three types of scenarios to run through when determining the agent status:

1. API is UP and agent is reporting UP via the OpenStack service (typical, returns 1).
2. API is UP and agent is reporting DOWN via the OpenStack service (typical, returns 0).
3. API is DOWN and the plugin is unable to gather a service list from nova (returns something like, "unable-to-reach-nova-api").

I know metrics have data types so I'm not sure if this is directly feasible without making a change to that value as well. The idea here is that we need to make it evident as an operations engineer who may be running this plugin locally on the node. If the API is down, it should notate that. In the above scenario, the templates would remain untouched, or we could add a warning message for this status. I'm actually thinking this may get us into trouble down the road if we decide to eventually create tickets based on warning messages. I'm thinking we may be better off not changing this side of things. The service-specific API check that we spoke to last week would bridge the gap of scenario 3 and only create a ticket if the API was to truly down.

Thoughts?

tonytan4ever:
@npawelek 
Let's look at scenario 3 a bit closer:

3. API is DOWN and the plugin is unable to gather a service list from nova (returns something like, "unable-to-reach-nova-api").

This is what exactly this change is intended to handle.  If the API is down (or the API hits 500/400, or other types of scenario), the code should hit: 

https://github.com/rcbops/rpc-maas/blob/master/playbooks/files/rax-maas/plugins/neutron_service_check.py#L35

And then with a `status_err` the plugin exits with (1).  So in this case, the metric value we generated
is `metric_bool('client_success', False, m_name='maas_neutron')`

We get one 'client_success' metric, which is 0 (False), and in our template,  we catch this template, generate a warning rather than letting it error out with a critical warning like this:

> 1 ["CRITICAL Plugin didn't finish in 59 seconds 1526500887466 UP"]

We catch it and let plugin finish by:

> if (metric["client_success"] != 1) {
>        return new AlarmStatus(WARNING, "neutron l3 agent can't reach API");

Let me know if this clarifies.

npawelek:
@tonytan4ever Correct, but we're not actually altering anything with this change, it essentially remains as-is. The idea was to fix the cascading fallout of API down related events, which I'm not sure we can achieve via a simple adjustment to the template.

tonytan4ever:
@npawelek:  Roger. Adjusted system calls to return 0, that should do it.

tonytan4ever:
recheck_xenial_pike_deploy

tonytan4ever:
recheck_xenial_newton_deploy

tonytan4ever:
recheck_xenial_pike_deploy

tonytan4ever:
Thanks @cloudnull 
Given current pike gate is not that important, and other gates are passing, merging this one.

PR #531, opened 2018-07-02
Support category/check/alarm doc descriptions

briancurtin:
To support the need for prose documentation alongside the automated check docs that are rendered, this change introduces a `stubs` directory that includes documentation stubs that are inserted during the build at the appropriate spot for said stub. Each category and check includes a "title.rst" stub file for category and check-wide descriptions to go under their respective titles, and alarm descriptions are named after the alarm.

This includes empty stubs for everything right not that makes a complete build. If stubs are missing for any category, check, or alarm, that will cause the build to emit a warning log, which at current will kill the build. If stubs are empty, which all of them are right now, the build will emit an info message to the log stating that it found no content.

### DIFF
--- a/doc/maas-checks/source/README.rst
+++ b/doc/maas-checks/source/README.rst
@@ -23,6 +23,10 @@ If any of the stub files are not found, a warning will be raised by the
 Sphinx build. If files are found but they are empty, an informational
 message will be logged to the build output.
 
+NOTE: Do not use =, -, or * as any potential section headers you create
+inside any of the stub files. Those are already in use for the structure
+of the rest of the document.
+
 maas_checks_config.py
 =====================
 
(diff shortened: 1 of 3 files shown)


briancurtin:
An example of what this ends up looking like: https://i.imgur.com/SipFNPU.png

I added the text you see there to `stubs/api/title.rst`, `stubs/api/lb_api_check_octavia/title.rst`, and `stubs/api/lb_api_check_octavia/lb_api_alarm_octavia.rst`. As that text gets pulled directly into the file, it can be formatted as reStructuredText in any way you want, though you will need to be careful if you're going to add your own section headings inside of any of those file so they don't conflict with what is currently being used elsewhere, which is `=`, `-`, and `*` which I just added to the README.

PR #534, opened 2018-07-03
Remove tox for Ceph jobs

andymcc:
The RPC-Ceph gate already has a funcitonal test that runs and deploys
MaaS. Utilizing tox from the MaaS side means these tests don't work the
same way and causes some issues. We can remove the tox portion and just
utilize the rpc-ceph job directly, which will then handle the
configuration and deployment of MaaS from the rpc-maas repo that was
cloned as part of the test run.


tonytan4ever:
LGTM,  ceph is passing,  the pike failure will be addressed in another PR

PR #545, opened 2018-08-08
Updated rally_diag.sh

shahzaib-bhatia:
Change query for ports to work more better and dress up all the queries.


tonytan4ever:
LGTM

PR #550, opened 2018-08-27
Turtles 1331

tonytan4ever:


npawelek:
There's no real need to start the agents as this is done already at the end of site.yml. This ensures that prematurely ending a deployment and restarting it will not potentially activate broken things. Additionally it ensures the agent is started on both RPCO/RPCR deployments.

releasenotes/notes/TURTLES-1331-enable-but-not-start-rax-mon-5b369888d36e9681.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    Enable but not rax-mon service in maas-agent-setup
```
hmakkapati: but not start?

tonytan4ever: See @npawelek's comment

PR #556, opened 2018-09-06
Fix bonding slave_count logic

npawelek:
Fix the bonding mii split logic to remove leading and trailing spaces and calculate the slave_count outside the loop determining if a slave is down.


npawelek:
recheck_container_bandit_tox-test

PR #558, opened 2018-09-13
JIRA: REO-336 Fix octavia playbook issues due to rpc-r additions

shannonmitchell:
https://github.com/rcbops/rpc-maas/blob/master/playbooks/maas-openstack-octavia.yml#L73-L92

  It looks like the 'Get project id (RPC-R)' task is overwriting the octavia_project_return
  registered variable even when skipped.  This causes the 'Set project id' task to fail on
  rpc-o environments as it tries to pull from the stdout of the shell command.  This fix just
  sets up a different var and set_fact task for each based on the same conditionals.


shannonmitchell:
recheck_container_bandit_tox-test

PR #564, opened 2018-10-01
Add deploy timestamp to metadata

hmakkapati:


playbooks/templates/common/macros.jinja
```diff
@@ -30,6 +30,6 @@ metadata          :
   osa_version: "{{ lookup('pipe', get_product_osa_version_cmd )}}"
 {% endif %}
   rpc_maas_version: "{{  lookup('pipe', 'cd .. && git describe --tags --abbrev=0') }}"
-  rpc_maas_deploy_date: "{{ ansible_date_time.date }}"
+  rpc_maas_deploy_date: "{{ ansible_date_time.iso8601 }}"
```
CrashenX: I support this change. I'll defer to @tonytan4ever to review. Thanks!

tonytan4ever:
recheck_trusty_newton_deploy

playbooks/templates/common/macros.jinja
npawelek: I suspect this will cause all templates to receive a different timestamp during a deployment. We initially discussed this during implementation, but didn't believe it offered any benefit. Is there something specific we're trying to achieve here?

hmakkapati: Ooh, good point. The intention here was to identify multiple deployments on the same day

npawelek: @tonytan4ever @hmakkapati I'm good with generating a fact at playbook runtime and using the same value across a given deployment. If ansible is cancelled and re-run or targeted, a new timestamp will be updated in those templates.

PR #565, opened 2018-10-01
Cephstora 453

tonytan4ever:


tonytan4ever:
recheck_trusty_mitaka_deploy
recheck_xenial_newton_deploy

tonytan4ever:
This will address: TURTLES-1181 and CEPHSTORA-359 as well.

playbooks/library/raxmon.py
```diff
@@ -139,15 +150,32 @@ def create_agent_token(module, conn, entity):
         module.fail_json(msg=msg)
 
 
+def delete_entity(module, conn, entity):
+    entities = _get_entities(conn, entity)
+    for entity in entities:
+        try:
+            conn.delete_entity(entity)
```
hmakkapati: Is the goal here to delete all entities or just the given entity. I thought it's the latter?

tests/test-raxmon-assign-agent-to-entity.yml
```diff
@@ -0,0 +1,31 @@
+---
```
hmakkapati: Where is this playbook used?

tonytan4ever: This one is not used in the regular playbook (site.yml),  really just here for testing just in case, we need to create an entity.  
Same for the delete entity play.

PR #566, opened 2018-10-03
K8S-1694 Remove k8s api checks

xgerman:
Since we moved the k8s api server checks to the k8s cluster maas,
they can be deleted here.


xgerman:
rebased but it still seems to fail some deploys

tonytan4ever:
recheck_xenial_pike_deploy
recheck_xenial_newton_deploy

PR #567, opened 2018-10-03
(ASC-897) Added tags to 'Run MaaS verify local' task in maas-verify.yml

phongdly:
ASC team would like to run check on local only to verify some openstack API calls


tonytan4ever:
recheck_xenial_newton_deploy

PR #548, opened 2018-08-21
RPC-R Ansible Maas

tonytan4ever:
First target was for RPC-R env

PR #568, opened 2018-10-03
Adding agent/poller installation support for RPC-R

tonytan4ever:


briancurtin:
recheck_all

tonytan4ever:
recheck_xenial_ceph_deploy

tonytan4ever:
recheck_xenial_newton_deploy

playbooks/maas-agent-setup.yml
```diff
@@ -41,9 +42,9 @@
       file:
         path: "{{ maas_plugin_dir }}"
         state: directory
-        owner: "root"
+        owner: "{{ ansible_ssh_user | default('root')}}"
         group: "root"
-        mode: "0755"
+        mode: "0757"
```
npawelek: Is there a reason this was changed from `0755` to `0757`? We shouldn't want these writeable by everyone.

tonytan4ever: This is what's there left in the feature branch:
https://github.com/rcbops/rpc-maas/pull/548/files#diff-91e2993f1af4f36c72349d76a45c789dR152

I vaguely recall it's that if we don't use 757,  we will run into a permission issue ?

tonytan4ever: Changed back to 755

PR #578, opened 2018-10-24
Report conntrack count for all namespaces

BjoernT:
In addition to checking the container root namespace,
nested namespaces inside the container are checked against the
container maximum configured connection tracking count.
Only the highest connection tracking count is reported, as
MAAS limites the number of metrics it can process.

Closes-Bug: TURTLES-1006


CrashenX:
@tonytan4ever  can haz review?

PR #599, opened 2018-12-05
Autoload OS_CACERT environment variable

tonytan4ever:


inventory/rpcr_dynamic_inventory.py
```diff
@@ -91,6 +92,16 @@ def load_stackrc_file(self):
                 v = match.group('value').strip('"').strip("'")
                 os.environ[k] = v
 
+    def load_os_cacert(self):
+        os_cert_path = '/etc/pki/ca-trust/source/anchors/'
+        if os.path.exists(os_cert_path):
+            # load the cert file in this directory
+            for f in os.listdir(os_cert_path):
+                f_path = os.path.join(os_cert_path, f)
```
npawelek: I believe this is concatenating instead of delimiting each file

tonytan4ever: This only sets **OS_CACERT** to the first file in that location, it's a equivalent of this:

```
[root@directorvm stack]# export OS_CACERT=$(find /etc/pki/ca-trust/source/anchors/ -type f -exec ls -1 {} \;)
[root@directorvm stack]# echo $OS_CACERT
/etc/pki/ca-trust/source/anchors/undercloud.crt.pem /etc/pki/ca-trust/source/anchors/overcloud.crt.pem
```
So Why would we want delimiting all the fies and set it to the **OS_CACERT** though ?

tonytan4ever:
recheck_ bionic_mnaio_3ctlr_2comp_3ceph_osp_13_deploy

tonytan4ever:
recheck_bionic_mnaio_3ctlr_2comp_3ceph_osp_13_deploy

PR #601, opened 2018-12-17
Add type to deploy_osp option in ceph_osd_host_facts module

npawelek:
This was preventing fact gathering from succeeding in non-OSP environments. It was causing the deploy_osp option to always be set.


tonytan4ever:

tonytan4ever:

PR #605, opened 2019-01-23
Removing pyrax dependency in testing

tonytan4ever:


tests/maasutils.py
```diff
@@ -1,38 +1,61 @@
 #!/usr/bin/env python
 import os
+import sys
 
 import click
-import pyrax
+from rackspace_monitoring.providers import get_driver
```
npawelek: I would opt to remove raxmon as a dependency here (lots of packages) and get rid of the unnecessary functions. We already have ansible module for obtaining the entity id (used in agent setup). At the present time, we only use this for generating a gating auth token and api url.

tonytan4ever: I'd opt to leave it there as respect of history, lol. 
set_webhood could be useful when testing in the future,  also raxmon's dependency is not that much really: https://github.com/racker/rackspace-monitoring/blob/master/setup.py#L174~L175

PR #607, opened 2019-01-24
Fix usage of OS_IMAGE_API_VERSION for glance in maas_common.py

npawelek:
The usage of OS_IMAGE_API_VERSION in maas_common.py was explicitly obtaining the value from environment variables. This change actually exports the values when maasrc is sourced from maas_common.py.


tonytan4ever:
recheck_trusty_liberty_deploy
recheck_xenial_pike_deploy
recheck_trusty_newton_deploy
recheck_xenial_rocky_deploy

PR #619, opened 2019-02-04
Bandit testing

racker-rick:
modified swift-recon.py so that get_container_name works for both lxc and docker without shell==True so that bandit tests pass
swift-recon.py was raising bandit high sev issues due to us self signing certs and allowing for the possibility of shell injection by running subprocesses with shell set to True.    It was only set up this way so that the container name could be retrieved from a docker ps  command - awking for the swift-proxy container.  In order to resolve this issue, the full list of docker issues was retrieved.  This list of containers was then searched for the entry (if it existed) containing the name swift-proxy.  If found, the container name was returned.  All without having to invoke a subprocess that used shell output.

Also, to resolve the self signed certs issue, we are skipping the bandit test, B501, that checks for this...as we know it will fail every time.

### DIFF
--- a/playbooks/files/rax-maas/plugins/swift-recon.py
+++ b/playbooks/files/rax-maas/plugins/swift-recon.py
@@ -72,7 +72,7 @@ def get_container_name(deploy_osp, for_ring):
         containers_list = subprocess.check_output(get_containers.split())
 
         c = getcontainer('swift_proxy', containers_list)
-        if c is not None:
+        if c:
             container = c.split()[-1]
 
     return container
--- a/playbooks/files/rax-maas/plugins/swift-recon.py
+++ b/playbooks/files/rax-maas/plugins/swift-recon.py
@@ -57,10 +57,8 @@ def get_container_name(deploy_osp, for_ring):
         # identify the container we will use for monitoring
         get_container = shlex.split(
             'lxc-ls -1 --running ".*(swift_proxy|swift)"')
-        shell = False
-
+        
         try:
-            #containers_list = subprocess.check_output(get_container, shell=shell)
             containers_list = subprocess.check_output(get_container)
             container = containers_list.splitlines()[0]
         except (IndexError, subprocess.CalledProcessError):
@@ -71,12 +69,9 @@ def get_container_name(deploy_osp, for_ring):
         get_containers = (
             "/usr/local/bin/docker ps -f status=running"
         )
-        shell = False
-        #containers_list = subprocess.check_output(get_containers.split(), shell=shell)
         containers_list = subprocess.check_output(get_containers.split())
 
         c = getcontainer('swift_proxy', containers_list)
-
         if c != None:
             container = c.split()[-1]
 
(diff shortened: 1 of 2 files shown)


playbooks/files/rax-maas/plugins/swift-recon.py
```diff
@@ -40,34 +40,50 @@ class CommandNotRecognized(maas_common.MaaSException):
     pass
 
 
+def getcontainer(pattern, containers):
+    c = None
+
+    for container in containers.split("\n"):
+      if re.search(pattern, container):
```
cloudnull: 4 space

playbooks/files/rax-maas/plugins/swift-recon.py
```diff
@@ -40,34 +40,50 @@ class CommandNotRecognized(maas_common.MaaSException):
...
+            "/usr/local/bin/docker ps -f status=running"
         )
-        shell = True
+        shell = False
+        #containers_list = subprocess.check_output(get_containers.split(), shell=shell)
+        containers_list = subprocess.check_output(get_containers.split())
+
+        c = getcontainer('swift_proxy', containers_list)
+
+        if c != None:
```
cloudnull: this can simple be:

``` python
c = getcontainer('swift_proxy', containers_list)
if c:
    ...
```

tox.ini
```diff
@@ -56,8 +56,8 @@ commands =
 # that exit status. Then run it again on the highest severity and confidence
 # levels and get our pass/fail status from that.
 # This is reported to bandit at https://github.com/PyCQA/bandit/issues/341
-    - bandit -s B404,B410 -i -l -r playbooks/
-    bandit -s B404,B410 -iii -lll -r playbooks/
+    - bandit -s B404,B410,B501 -i -l -r playbooks/
```
cloudnull: It would be good to capture why we're skipping these codes.

playbooks/files/rax-maas/plugins/swift-recon.py
```diff
@@ -40,34 +40,50 @@ class CommandNotRecognized(maas_common.MaaSException):
     pass
 
 
+def getcontainer(pattern, containers):
```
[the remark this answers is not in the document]
racker-rick: I wasn't sure about this as container=None initially.  Also wasn't sure if you were guaranteed to find the search pattern in the list of containers.  I was simply preserving the context of the original code.  Is the state/context still None after iteration through containers?  I figured safe rather than sorry?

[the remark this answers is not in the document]
cloudnull: It's functionally the same thing, with the existing code, if there are no matches this function still returns `None`.

tox.ini
racker-rick: added comments regarding skip of B501

BjoernT:
To my knowledge Bandit has been replaced with Checkmarx internally, not sure if we want to run maas twice between different AST tools

racker-rick:

PR #621, opened 2019-02-06
Remove remnant tigk configuration

npawelek:
This change removes remnant tigk configuration that was missed
during the initial removal.


npawelek:
recheck_xenial_pike_deploy

PR #626, opened 2019-02-14
Fixing OSP gates

tonytan4ever:
A few variable and file passing mechanism change to fix OSP gate

Signed-off-by: tonytan4ever <tony.tan@rackspace.com>

### DIFF
--- a/playbooks/maas-pre-flight.yml
+++ b/playbooks/maas-pre-flight.yml
@@ -12,62 +12,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-- name: Check for RHOSP
-  hosts: localhost
-  gather_facts: false
-  become: true
-  tasks:
-    - name: Check for /etc/rhosp-release
-      stat:
-        path: /etc/rhosp-release
-      register: osp_release_file
-
-    - name: Check for /usr/bin/tripleo-ansible-inventory
-      stat:
-        path: /usr/bin/tripleo-ansible-inventory
-      register: tripleo_inventory_file
-
-    - name: Set deploy_osp fact
-      set_fact:
-        deploy_osp: "{{ (osp_release_file.stat.exists and tripleo_inventory_file.stat.exists) | bool }}"
-
-  tags:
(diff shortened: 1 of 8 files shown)


tonytan4ever:
recheck_xenial_ceph_deploy

playbooks/templates/rax-maas/run_plugin_in_venv.sh.j2
```diff
@@ -4,6 +4,10 @@ if [[ -f {{ maas_openrc | default('~/openrc') }} ]]; then
   source {{ maas_openrc | default('~/openrc') }}
 fi
 
+if [[ -f /home/stack/stackrc ]]; then
```
npawelek: This will only exist on the director node and it should not be pertinent to overcloud which plugins are targeting.

tonytan4ever: this is resolved

playbooks/maas-pre-flight.yml
```diff
@@ -12,30 +12,40 @@
...
-        path: /usr/bin/tripleo-ansible-inventory
-      register: tripleo_inventory_file
+    - name: Remove local maas.facts
+      file:
+        dest: "/etc/ansible/facts.d/maas.fact"
+        state: absent
 
-    - name: Set deploy_osp fact
-      set_fact:
-        deploy_osp: "{{ (osp_release_file.stat.exists and tripleo_inventory_file.stat.exists) | bool }}"
```
npawelek: This is actually needed because the OSP vars task littered through the plays explicitly look for the fact over the variable with the dynamic generation changes. This was done to begin the work to remove the need for the OSP vars to be included in every play.

tonytan4ever: I understood,  but deploy_osp isn't working consistently throughout my testing so I moved deploy_osp into rpcr_dynamic inventory, thus this is no longer needed.  Technically rpcr_dynamic inventory should be able to figure out all OSP vars in the future.
I did put rhosp release logic later in preflight code and that part is needed.
IMO using set fact is not as direct and not working well in my tests so I had to move it away.

tonytan4ever:

tonytan4ever:

PR #633, opened 2019-02-21
NATT-73: Remove OSP vars and tasks

tonytan4ever:


tests/user_osp13_vars.yml
```diff
@@ -27,3 +27,8 @@ maas_env_identifier: "test"
 # Disable remote.http checks because AIO endpoints will always be private
 # addresses. Private network monitoring will need to be added.
 maas_remote_check: false
+
```
npawelek: I suspect we'll need these in customer environments as well? I don't think we can put them in in any of the standard vars files, but we could add them to the `maas_compat` dictionary in `playbooks/vars/maas.yml`. This way they're included as a local fact in OSP deployments.. we would just need to update the required templates to pull in the local fact.

tonytan4ever: I looked over the repo and there's no templates or other places that's using these variables/facts.

We shouldn't need them.  Putting it in user vars is really just my temporary measure to sort out what are the all variables needed in templates/plays.  All other variables will be provided in the dynamic inventory python script.  In future if we have any osp related variables we will use dynamic inventory to provide as well.

tonytan4ever:
recheck_all

tonytan4ever:
recheck_xenial_queens_deploy

tonytan4ever:
recheck_xenial_queens_deploy

tonytan4ever:

tonytan4ever:

tonytan4ever:

PR #635, opened 2019-02-22
NATT-83: Add network statistic error rate check

npawelek:
This check gathers physical network interface receive and transmit errors, sums
all error values, and evaluates based on a rate-over-time function in MaaS. By
default it looks for a total greater than 64 over a 60s period of time.


npawelek:
recheck_trusty_liberty_deploy

PR #643, opened 2019-03-20
Remove keystone user create for maas

npawelek:
Removing as this does not appear to be used any longer.


npawelek:

PR #646, opened 2019-03-28
Use OSA version to detect holland virtualenv

npawelek:
In the event that OSA and RPCO versions are not identical, this change
uses the OSA version to detect whether or not holland is deployed into a
virtualenv.


npawelek:

npawelek:
recheck_all

PR #648, opened 2019-04-01
Fix Kilo endpoint detection and venv pip options

npawelek:
The '--format value' option did not exist in Kilo. This change uses csv
format to identify the keystone endpoint, which also exists in later
versions. Also provides an interface override in-case the internal
interface is v2.


npawelek:
recheck_bionic_mnaio_osp13_deplo

npawelek:

PR #649, opened 2019-04-02
Enable release-specific disabled checks for RPCR OSP

npawelek:
This change automatically disables known release-specific checks during
an OSP deployment.


npawelek:
recheck_all

PR #651, opened 2019-04-18
Create additional check name for container_conntrack_count

npawelek:
This change fixes the conntrack_status check run in the
maas-container-conntrack.yml playbook to prevent overwriting the
physical conntrack_count check. This will create two separate checks with
different names off the same template.


npawelek:
recheck_xenial_queens_deploy

PR #655, opened 2019-05-14
Refactor ceph osd alarms

npawelek:
This change refactors the ceph_osd_stats checks to have a 1:1
relationship of checks to alarms. The refactor utilizes the local socket
versus using expensive ceph API calls. Playbooks will also
attempt to remove the legacy osd_stats_check templates including those
with stale OSD IDs.

### DIFF
--- a/playbooks/files/rax-maas/plugins/ceph_monitoring.py
+++ b/playbooks/files/rax-maas/plugins/ceph_monitoring.py
@@ -66,20 +66,13 @@ def get_ceph_status(client, keyring, fmt='json', container_name=None,
                          deploy_osp=deploy_osp)
 
 
-def get_ceph_pg_dump_osds(client, keyring, fmt='json', container_name=None,
-                          deploy_osp=False):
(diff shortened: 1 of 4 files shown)


npawelek:
recheck_xenial_queens_deploy

npawelek:
The existing methodology of calling `ceph osd dump` would be too expensive for the 1:1 refactor. As a result we'll need to take a step back and utilize the local socket on the OSD nodes to detect if each OSD ID is actually up. Something like `ceph daemon osd.<id> status` would provide the OSD status and shouldn't cause too many issues as the call is handled via the local socket and not via the mons nodes.

This functionality works with jewel and luminous. Testing to confirm if the same method is functional with mimic and nautilus because making the necessary changes. Additionally, filesystem checks should also be removed/added when ID's change so this will need to be worked in as well.

npawelek:
recheck_xenial_newton_deploy

npawelek:
recheck_xenial_newton_deploy

npawelek:

PR #659, opened 2019-05-24
Move templates older than deploy date

npawelek:
This change will find check templates that are older than the deployment
date plus 8 hours and move them to a backup directory. This helps to
ensure only checks that should be deployed, are deployed.


npawelek:
recheck_trusty_mitaka_deploy

PR #660, opened 2019-05-24
Bump embedded Ansible version to 2.7.11

npawelek:
The most notable change upgrading Ansible from 2.6.12 to 2.7.11 is that
Ansible 2.7+ requires setuptools in order to build an isolated
virtualenv with --no-system-site-packages (default). Our typical OSA
deployments alternatively replace the --index-url with that of the repo
container for all pip related tasks. As setuptools is now required, but
not part of the repo, it will fail. These changes will temporarily
remove the index-url from the pip.conf to successfully build the
virtualenv.

### DIFF
--- a/playbooks/common-tasks/maas-infra-ubuntu-install.yml
+++ b/playbooks/common-tasks/maas-infra-ubuntu-install.yml
@@ -24,9 +24,8 @@
 
 - name: Ensure galera packages are installed
   apt:
-    pkg: "{{ item }}"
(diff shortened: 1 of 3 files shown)


npawelek:
Dropping 2.8.0 to 2.7.11. The inventory for all groups will need to be modified in 2.8 as the `-` (dash) is no longer valid for group names. This will require extensive changes and testing to ensure everything remains functional.

npawelek:

PR #663, opened 2019-05-30
Simplify maas_proxy_url logic

npawelek:
This changes removes usage of HTTP as the endpoints don't communicate
using HTTP. Instead, when a proxy is defined, we opt to use openssl
directly to determine whether the connection is possible without a
proxy. If this fails it will fallback to defining the maas_proxy_url
variable.


npawelek:
recheck_trusty_newton_deploy

PR #669, opened 2019-06-12
Remove octavia_process_names_sanitized and only use octavia_process_n…

BjoernT:
…ames

The process name on OSA builds do not add a log to the worker
process, hence the sanitized dict is no longer required


npawelek:
recheck_container_releasenotes_tox-test

PR #676, opened 2019-09-16
Add introspection to account for local rabbitmq management address

npawelek:
This change adds some basic introspection to the maas-infra-rabbitmq.yml
playbooks. The latest minor release of OSP13 modified the rabbitmq.config
management section to lock down access explicitly to localhost. Dynamic
inventory cannot be updated as it would break prior releases, and puppet
can't effectively be modified to prevent changes in later OSP releases from
breaking it again.


npawelek:
recheck_trusty_newton_deploy
recheck_xenial_ceph_deploy
recheck_xenial_newton_deploy

npawelek:
recheck_trusty_newton_deploy
recheck_xenial_ceph_deploy
recheck_xenial_newton_deploy

npawelek:
this only affects osp13 which listens exclusively on localhost in later versions, if it's not there, it'll fallback to the same way it's always worked

PR #680, opened 2019-09-28
changed hp_monitoring get_hpasmcli_status back to get_chassis_status

sdhardy:
In commit 99ee251a4c5c7a60092766d3b9556b69c8cce5a1, function get_hpasmcli_status() was used but it wasn't defined. I changed references to this back to get_chassis_status to get the script working again.


npawelek:
@sdhardy nice catch!

PR #701, opened 2020-08-28
fixing ceph_monitoring.py for nautilus

Alfano93:
ceph_health_status = ceph_status['health']['overall_status'] is no longer available through ceph status. Since this value does not exist, the plug-in crashes when run. Since we no longer have to check for any hammer or jewel clusters, removing the if statement is the best option to fix this.

### DIFF
--- a/playbooks/files/rax-maas/plugins/ceph_monitoring.py
+++ b/playbooks/files/rax-maas/plugins/ceph_monitoring.py
@@ -120,11 +120,16 @@ def get_cluster_statistics(client=None, keyring=None, container_name=None,
                                   deploy_osp=deploy_osp)
     # Get overall cluster health
     # For luminous+ this is the ceph_status.health.status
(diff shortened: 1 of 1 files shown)


playbooks/files/rax-maas/plugins/ceph_monitoring.py
```diff
@@ -120,14 +120,16 @@ def get_cluster_statistics(client=None, keyring=None, container_name=None,
...
+    try:
+      ceph_health_status = ceph_status['health']['overall_status']
+    except KeyError:
+      ceph_health_status = ceph_status['health']['status']
+
     metrics.append({
-        'name': 'cluster_health',
-        'type': 'uint32',
-        'value': STATUSES[ceph_health_status]})
+    'name': 'cluster_health',
```
BjoernT, line 130: Indentation is now off

BjoernT:
See #702

PR #702, opened 2020-08-28
Fixing health check for nautilus

BjoernT:
ceph_health_status = ceph_status['health']['overall_status']
is no longer available through ceph status.


shannonmitchell:
Just a heads up.  I'm fixing and moving gating over to github actions.  The ones in place should be working with more to come.

PR #704, opened 2020-09-01
JIRA: REPCOS-44 Ignore some ceph health checks

shannonmitchell:
After a meeting with the storage team, it looks like a
quick fix to ignore certain ceph health checks would be
helpful until we can break out things into individual
checks and alarms per type of breakage.


shannonmitchell:
This was tested in production yesterday.

PR #714, opened 2020-10-14
Remove cinder quota check from octavia plugin

BjoernT:
Octavia does not consume cinder volumes and
the volume check can be removed.
Apart from the issue that the quota check is not scoped to the
actual service tenant, resulting into false alerts.


shannonmitchell:
I ran the full set of tests for train and queen on the devices that failed using the same commands.  The verification passes without issue.  

The gating errors look like a possible timing issue with the raxmon module when creating the entity and agent id.  The 2nd run works after the entity has been created.

shannonmitchell (approved):
looks good to me

PR #725, opened 2021-03-02
Make storage check use df in container namespace

### LINKED ISSUE #724: Container storage check plugin fails when /etc/mtab is a symlink
The current [container storage check](https://github.com/rcbops/rpc-maas/blob/a3f9183fea6ff9f354df325fef324ae596fbdf26/playbooks/files/rax-maas/plugins/container_storage_check.py) does a chroot into `/proc/<container init process pid>/root` and calls `psutil.disk_partitions()`. This fails when `/etc/mtab` is a symlink to `../proc/self/mounts`, which is the default configuration in current linux distributions.

```
OSError: [Errno 2] No such file or directory: '/proc/self'
```

This happens because the container is running in a separate pid and mount namespace and /proc/self is not present in the chroot.

Using nsenter with pid and mount namespace works correctly and shows the correct mounts for the container (in this example init_pid is 5169):

```
nsenter -m -p -t 5169 -- /bin/bash -c 'cat /etc/mtab'
```

It may be possible to achieve the same thing from python with something like [python-nsenter](https://github.com/zalando/python-nsenter). Another option would be to use the `lxc.attach_wait` method to run a command like `df`.

JCallicoat:
The previous method of using chroot to check container storage fails
when the container uses a symlink from /etc/mtab to ../proc/self/mounts
since the container is running in an isolated pid/mnt namespace and
/proc/self is not visible in the chroot.

This commit uses the the `attach_wait` method from the `lxc` module to
run the df command in the container pid/mnt namespace and parses the
df output to gather usage percentage for each mounted device.

Fixes #724

### DIFF
--- a/playbooks/files/rax-maas/plugins/container_storage_check.py
+++ b/playbooks/files/rax-maas/plugins/container_storage_check.py
@@ -16,7 +16,6 @@
 
 import argparse
 import os
-import psutil
 
 import lxc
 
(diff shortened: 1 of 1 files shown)


BjoernT:
Yes that's how we had it pretty much done for other checks to use lxc-attach.
But these checks had been deactivated due to pytho3 issues with the lxc-python module. Did you test this on a recent deployment under python3?

rax-rstark:
Is it worth having a python2 branch with this fix in place?

BjoernT:
> Is it worth having a python2 branch with this fix in place?

We already have a python2 branch in place but we need to check if this was also used for docker containers (OSP) or just lxc

PR #740, opened 2022-05-16
Check if python3 is installed in the ansible venv

sdhardy:
The script is checking if the ansible_venv directory exists. We have cases where the venv was present but was created with python2. With this change, the ansible_venv will be recreated if python3 isn't installed.


BjoernT:
OK i think there needs more to be done but this is ok

